Issue #176, opened 2026-09-05
Julia: a DEAD observed is dropped from BuildInspection.observed_exprs, so a §6.6.5 assertion on it is unanswerable


ctessum-claude:
## Summary

In the Julia binding, an OBSERVED that no live equation consumes — the ordinary
shape of a *diagnostic* an inline test wants to assert — never reaches
`BuildInspection.observed_exprs`. `_observed_field` (`src/pde_inline_tests.jl`)
resolves the asserted variable's body out of that map, so the assertion fails
with

```
array state 'g' has no cells in var_map
```

even though the document declares `g`'s defining equation and both the Rust and
the Python runners answer it.

This is separate from — and survives — the state-dependent-array-observed fix in
PR (this branch): with a live consumer the Julia runner now answers a
state-dependent array observed correctly; it is the *deadness* that is left.

## Reproduction

```json
{"lhs": "live", "rhs": {"op": "*", "args": [2.0, "u"]}}
```

on a document whose only ODE is `D(u,t) = 0*u`, with

```json
{"variable": "live", "time": 0.0, "reduce": "max", "expected": 8.0}
```

* Julia — `ERROR ... array state 'live' has no cells in var_map`
* Rust — PASS
* Python — PASS

The state-FREE sibling in the same document (`dead[i] = 3*i`, also consumed by
nothing) **does** pass in Julia, so the gap is specifically an observed that is
both dead and not build-materializable.

Verified again on the observed of a live document: making the same `g` live (its
value drives a second state) makes Julia answer it, which is the workaround the
new `tests/conformance/pde_inline_observed_state_dependent/` fixture uses so the
category can require all three bindings.

## Why it matters

An inline test's natural target is a quantity computed *for the test* — a
tendency, a flux, a diagnostic — which by construction nothing else reads. On
Julia such a variable is assertable only if the author also wires it into the
dynamics, which changes the model to make the test runnable.

## Suggested fix

When the build's observed graph has no entry for the asserted name, fall back to
the component's own defining equation from the loaded `EsmFile`, evaluated in
the same scope `_observed_field` already builds (build constants + resolved
parameters + — since the state-dependent fix — the trajectory sample), with the
component's other observeds resolvable by name. Rust reaches the same place by
REQUESTING the observed from the runtime (`SolveOptions::output_observed`),
which sidesteps the question of whether anything else reads it.

Filed from EqWeFiC (WRF-in-.esm); repro files can be supplied on request.

Issue #175, opened 2026-09-05
Julia tree-walk: an elementwise-defined array observed reached only through an aggregate gather fails with E_TREEWALK_UNBOUND_VARIABLE


ctessum-claude:
## Summary

In the Julia binding, an array-shaped **observed** whose body is written *elementwise* over another array — `f = 1 + cos(pi*zc)` with `zc` shaped `[lev]` — and which is then consumed **only** through an `index(f, j)` gather inside an `aggregate` body raises

```
E_TREEWALK_UNBOUND_VARIABLE: <Model>.zc
```

Rust evaluates the same document correctly, so this is a Julia-side gap rather than a document defect.

## Reproduction

A column document (the EqWeFiC `integral_column_midpoint.esm` convergence suite) with, per model:

```json
{"lhs": "f", "rhs": {"op": "+", "args": [1, {"op": "cos", "args": [{"op": "*", "args": ["pi", "zc"]}]}]}}
```

`zc` is shaped `[lev]`; `f` is shaped `[lev]`; `f` is read only as `index(f, j)` inside an `aggregate` (a cumulative column integral produced by an `integral` rewrite rule). Julia raises the unbound-variable error on `zc`; Rust returns the right numbers.

Two variations that DO work, which localize it:

1. Respelling `f` as an explicit gather — `f = aggregate(i from lev; 1 + cos(pi*index(zc, i)))` — makes Julia pass.
2. The same elementwise spelling consumed **directly in a state RHS** rather than through an aggregate gather (a forcing term `S = exp(-t)*(… cos(pi*zc) …)` in a heat-column problem) also works in Julia.

So the gap is specific to an *elementwise-defined array observed that is reached only via an aggregate gather*.

## Where it looks like it comes from

`pkg/EarthSciAST.jl/src/shape_promotion.jl` (~line 489):

```julia
# Names whose shape GREW in this pass (scalar → array). Only these need
# consumer-side gathers below: a pre-existing array operand was already the
# author's (or an earlier lift's) responsibility to index, and rewriting it
# here would change working documents.
newly_promoted = Set{String}(k for k in arrayvars if flat_was_scalar(flat, k))
```

[...]

Issue #174, opened 2026-09-05
esm-spec §9.7.7: import renaming does not rewrite an `integral`'s `var`/bounds, so an integral rewrite rule cannot be renamed onto another axis


ctessum-claude:
## Summary

§9.7.7 ("Transitivity (normative occurrence sites)") specifies that renaming an **index set** at an import edge rewrites

> the registry key; `of` parent lists of ragged/derived index-set definitions; `{"from": <name>}` references wherever the form is admitted; and the **`wrt`/`dim` axis scalar fields** of Expression nodes in template `body` **and `match`**

That list is exhaustive in the implementations, and it omits the axis-naming scalar fields of `integral`: `var`, and the bare axis names that appear as `lower` / `upper` bounds. As a result a **`D` rewrite rule** renames onto another axis exactly as intended (`D(f, wrt: "x")` imported under `rename: {"x": "lev"}` becomes a rule matching `D(f, wrt: "lev")`), while an otherwise identical **`integral` rewrite rule** does not: its `match` keeps `var: "x"`, no longer matches the consumer's `integral(..., var: "lev", ...)`, and the integral is left **unlowered**.

## Reproduction

Two documents, identical except for the axis name written inside the `integral` node; both import the same rule library under `rename: {"x": "lev"}`.

* `probe_integral_lev.esm` — the consumer writes the integral with `"var": "lev"`, `"upper": "lev"` (i.e. in the *renamed* vocabulary, which is the only vocabulary the consumer knows). The rule instance still matches on `var: "x"`, nothing fires, the `integral` survives lowering and the document does not evaluate.
* `probe_integral_lev_varx.esm` — same rename, but the consumer writes `"var": "x"`, `"upper": "x"` (the library's pre-rename name, which is invisible to a consumer that only ever named `lev`). This one lowers and runs.

The rule library (`probe_integral_rules.esm`) declares templates whose `match` is

```json
{"op": "integral", "args": ["f"], "var": "x", "lower": "a", "upper": "x"}
```

and whose `where` constrains `f` to `shape: ["x"]` (that half *is* renamed).

The asymmetry is exactly the missing occurrence site: the shape constraint and `{"from": "x"}` ranges in the body are rewritten to `lev`, the `wrt` of a `D` would be, but `var`/`upper` are not.

## Why it matters

[...]

PR #177, opened 2026-09-05
fix(bindings): a §6.6.5 assertion may read an ARRAY OBSERVED that moves with the state

ctessum-claude:
### 1. A §6.6.5 assertion may read an ARRAY OBSERVED that moves with the state

esm-spec §6.6.5 admits **any shaped variable** in a `coords` / `reduce`
assertion. Nothing in it restricts the target to a state, and nothing restricts
it to a state-**free** observed — §5.23's "a reference denotes its expansion" is
what makes an observed assertable at all, and it does not stop at rank 0. All
three executing bindings nevertheless answered

```
array state 'g' has no cells in var_map
```

because each looked in exactly two places: the ODE state rows, and the
build-time products. A state-dependent array observed is in **neither, by
construction** — only STATE-FREE observeds hoist into a build inspection's setup
arrays (their value is constant along the trajectory; a state-dependent one's is
not), and the output-node observed reconstruction exposes only **scalar**
observeds as trajectory rows.

The practical consequence: the natural target of a column model's inline test —
`dudt = D(D(u,lev),lev)` after lowering, or a plain `g = 2*u` — was unassertable
in every binding, while the same quantity as a 0-D observed was fine.

Each binding closes it on its own official pathway; none grew a shadow
evaluator:

| Binding | How |
|---|---|
| **Rust** | the runner REQUESTS the asserted array observed (`SolveOptions::output_observed`), which the array runtime already knew how to emit as one row per cell (`append_observed_trajectories`); the assertion then reads ordinary cell rows. `pointwise_observed_requests` → `assertion_observed_requests`, matching the declared rank to the assertion form. |
| **Python** | `simulation_array.observed_at_state` seeds an `EvalContext` with the build's once-materialized state-free products and replays `_materialize_observeds` over the dependency-ordered time-varying observeds at that `(t, y)` — the same driver, the same context shape, and the same numbers the per-step RHS saw. `SimulatedStates` now carries the built `EsmProblem` so the runner can reach the build. |
| **Julia** | `_state_scope` re-assembles the solved state (and `t`) from the flat state vector and puts it into the two scopes `evaluate_cellwise` already resolves against; without it the resolved observed expression raised `E_TREEWALK_UNBOUND_VARIABLE: <Model>.u`. |

A state-FREE array observed keeps its existing answer (the build-materialized
field). The two agree by construction, since a state-free body is constant along
the trajectory — and the new fixture asserts one of each so that stays pinned.

### 2. An array field belongs to ONE component (Python)

`state_cells` was a single pass that **unioned** the model-qualified stem match
and the bare-suffix match. A document with four components each declaring `w[x]`
therefore produced **four** cells at index `[1]`: a `coords` sample read
whichever component sorted first — always the same wrong one, silently — a
`reduce` collapsed over all four components at once, and a per-cell `reference`
then indexed past the end of the field (`IndexError`).

Julia's `_state_cells` and Rust's `state_cells` were already two-pass (exact
qualified/bare stem first, bare-suffix fallback only when no exact stem exists);
Python is now too, matching its own `_scalar_slot`.

This is what made a four-model repro disagree with Rust on every model but the
first, and what made three of five models of a column-integral convergence suite
report wrong L2 errors (0.70 and 0.99 against a `< 1e-4` expectation) or crash
with `IndexError: index 64 is out of bounds for axis 0 with size 64`. Both files
now agree with Rust model-for-model.

## What is in the change

* **New conformance category** `tests/conformance/pde_inline_observed_state_dependent/`
  — fixture, Julia-minted goldens, and per-binding adapters for Julia / Python /
  Rust, each gating every assertion against **both** the golden actual and the
  fixture's own declared `expected`. The fixture's right-hand side is a per-cell
  constant, so `u(t) = i² + i·t` is integrated exactly by every pinned solver
  family and the goldens are integrator-independent to machine precision: a
  divergence here is a semantics divergence, never an integrator one.
* **CONFORMANCE_SPEC §5.24** (+ §5.24.1 for the one-component rule) and
  **ESM_COMPLIANCE_VALIDATION_MATRIX** rows BEHAV-06-B-009 / BEHAV-06-B-010.
* The Rust unit test that pinned the OLD limitation
  (`state_free_array_observed_assertions_evaluate_directly`, which asserted that
  a state-dependent observed *must* error) becomes
  `array_observed_assertions_evaluate_directly` and pins the new contract,
  including that a name the component does not declare still errors. A Python
  twin covers the same three cases.

## Test results

* **Rust** — `cargo test --release`: 425 passed, 0 failed across 60 test
  binaries (the one pre-existing red, the old-contract unit test above, is
  rewritten in this change). New adapter: `1 passed`.
* **Python** — `pytest tests/`: 3145 passed before, unchanged after, with the
  same 19 pre-existing failures + 3 errors in
  `test_esio_provider.py` / `test_loader_ingest_and_select.py` /
  `test_loader_unit_conversion_conformance.py` (verified identical on a clean
  `a1dc9bb30` checkout — an environment gap, not this change). New adapter:
  `2 passed`; `test_pde_inline_tests.py`: `22 passed`.
* **Julia** — new adapter
  `conformance_pde_inline_observed_state_dependent_test.jl`: `52 passed / 52`.
* `ruff check` clean on every touched Python file; `cargo fmt --check` and
  `cargo clippy` report no new findings (the pre-existing set is byte-identical
  before and after).

## Known residual, tracked separately

The Julia build drops a **dead** observed — one no live equation consumes — from
`BuildInspection.observed_exprs`, so Julia cannot answer an assertion on it
whether or not it is state-dependent, while Rust (which requests it from the
runtime) and Python (which evaluates the ordered graph on demand) both can. That
is #176, and it is deliberately not what this category pins: the fixture's `g` is
a live node so the category can require all three bindings.

Two further gaps found alongside this work and filed rather than fixed here:
#174 (§9.7.7 import renaming does not rewrite an `integral`'s `var`/bounds, so an
integral rewrite rule cannot be renamed onto another axis) and #175 (Julia
tree-walk: an elementwise-defined array observed reached only through an
aggregate gather fails with `E_TREEWALK_UNBOUND_VARIABLE`).

https://claude.ai/code/session_011sw1myoKbfFvhrTLhnBza4


ctessum:
Two notes from wiring EarthSciModels' inline-test gate onto this branch (EarthSciML/EarthSciModels#8). Both are about `_observed_sample`, and both are cheaper to fix here than after this merges.

## 1. `_observed_sample` is reachable only through `SimulatedStates`, so a second caller has to copy it

The gate in EarthSciModels drives `esm_problem` / `solve` directly — per-file subprocess for OOM isolation, `cse=False`, tolerances pinned at `1e-10`/`1e-12` — so it holds an `EsmProblem`, not a `SimulatedStates`. `_observed_sample` takes the latter and reaches the build through `getattr(sim, "problem", None)`, so the gate could not call it and ended up with a near-verbatim twin: same `build`/`flat` extraction, same `(f"{model}.{variable}", variable)` loop, same `ndim == 0 → None`.

That is two copies of one resolution order, and they have already diverged (see point 2). Since everything after the first line is `prob`-only, the smallest fix is to take the problem rather than the run:

```python
def _observed_sample(prob, model, variable, state, t) -> np.ndarray | None:
    build = getattr(prob, "build", None)
    ...
```

with the one in-tree call site becoming `_observed_sample(sim.problem, ...)`. Exporting it as `observed_field_at(prob, name, t, state)` alongside `observed_field(prob, name)` would be even better — the pair would then cover the state-free and state-dependent halves of §6.6.5 under one public spelling, and the gate could drop its copy entirely. Happy to send that as a PR against this branch if you'd prefer.

## 2. `_observed_sample`'s nearest-node sampling is safe only because of how `simulate_states` calls it

`_evaluate_assertion` picks the sample by `argmin(|result.t - t|)`. That is exact today only because `simulate_states` guards it: it refuses a nearest node farther than `_SAVEAT_MATCH_TOL * max(1, |t|)` from the requested time. The guard lives in the *caller*, not in `_observed_sample`, so any caller that does not reproduce it gets silent off-time sampling instead of a refusal.

That is exactly what happened downstream. The gate passes no `saveat`, so `res.t` is the dense default grid (`np.linspace(t0, t1, DENSE_OUTPUT_MIN_POINTS)` unioned with the solver's native steps) and the nearest node sits up to `(t1-t0)/2/(len(res.t)-1)` away. Measured on this branch's own fixture with the assertion moved off the grid — `t=0.50005` on span `(0,1)`, node spacing `1e-4` — `reduce=mean` on the state-dependent `g` reports `20.0` (the value at the `t=0.5` node) where the true value at the asserted time is `20.00025`. Against the §6.6.4 default of `rel=1e-6` that is ~12× the tolerance budget.

Worth noting the fixture cannot catch this itself: its assertion times (`0.0`, `0.5` on span `(0,1)`) land exactly on the linspace grid, so nearest-node and interpolate-at-`t` agree to machine precision. The category passes either way.

Downstream I fixed it by interpolating the state vector at exactly `a.time` before the replay, matching what the array-**state** branch already does. If `_observed_sample` grows a `prob` signature per point 1, it may be worth moving the time-selection decision inside it — either carrying the `_SAVEAT_MATCH_TOL` guard so an off-grid caller gets a refusal rather than a wrong number, or interpolating — so the contract does not depend on each caller remembering.

Neither point blocks this branch as it stands; the semantics fix is right and the fixture pins it. They are about keeping the second binding-adjacent caller from drifting.

---

Issue #178, opened 2026-09-05
esm-spec §6.6 / schema: a shaped parameter's `default`, `parameter_overrides`, and `initial_conditions` are number-only, and an `ic` cannot read a state-free array observed — column tests cannot supply their inputs


ctessum-claude:
## Summary

A §6.6 inline test cannot supply a **shaped** input. `parameter_overrides` and `initial_conditions` values, and a shaped parameter's `default`, are `number`-only in `esm-schema.json`, and an `ic` whose right-hand side is a state-free array observed (a `const` gather) is rejected at build time. The only escape today is a test-injected rewrite-rule library (§9.7.10) whose rules lower a rewrite-target op `input_<x>` to an inline `const` column — one generated `.esm` per test, 10 columns each — which is what the EqWeFiC YSU column component had to do for all four of its regimes (`components/atmospheric_dynamics/ysu/tests/scm_call*_inputs.esm` in EarthSciML-adjacent repo `ctessum/EqWeFiC`).

This gates every column-physics component test that replays a Fortran kernel dump: the inputs *are* columns (θ, q_v, u, v, p, dz, K profiles), and the instantaneous-derivative test shape (observed tendencies asserted at `time: 0`) needs them as `parameter_overrides` / `initial_conditions` of one shared model, not as one model per regime.

## What fails (Rust `esm` from `inline-test-array-observeds` @ 17cac5d8d and Python on the same branch agree)

| spelling | Rust | Python |
|---|---|---|
| `"default": [1.0, 2.0, 3.0, 4.0]` on a parameter with `"shape": ["lev"]` | schema error (`oneOf`) | `[...] is not of type 'number'` (`$defs/ModelVariable.default`) |
| `"parameter_overrides": {"theta0": [ ... ]}` | schema error | `[...] is not of type 'number'` (`$defs/Test.parameter_overrides.additionalProperties`) |
| `"initial_conditions": {"u": [ ... ]}` | schema error | same (`$defs/Test.initial_conditions.additionalProperties`) |
| `ic(u) ~ theta0` where `theta0` is a state-free array observed defined by a `const` gather | `Invalid field initial condition ... neither a provider-served loaded field, a constant, nor a per-cell coordinate expression` | `SimulationError: ic(P.u): RHS is neither ...` |

Repro files: `probe_spikeA.esm`, `probe_v2.esm`, `probe_v4.esm` (attached below in reduced form).

```json
{
[...]

PR #179, opened 2026-09-05
fix(bindings): an inline `reference` binds the field's dimension names; override keys resolve to the longest dotted suffix (§6.6.5, §6.6.2, §4.6)

ctessum-claude:
## Summary

Two cross-binding fixes to the §6.6 inline-test runners, found while writing the first column-physics component tests (a WRF YSU boundary-layer scheme on a `lev` column, EqWeFiC Spike A):

1. **An inline `reference` binds the field's dimension names (esm-spec §6.6.5).** The spec says a `reference` is "an Expression whose free variables are the domain dimension names", but every executing binding evaluated it as one build-time array expression and sampled it per cell, so a dimension name mentioned free was unbound — Rust `E_TREEWALK_UNBOUND_NAME: 'lev'`, Python `Unresolved symbol`, Julia `E_TREEWALK_UNBOUND_VARIABLE` — and every reference had to be an explicit `aggregate(i from lev; …)` gather. `bind_dimension_names` (same name and mechanism in Julia, Python, Rust) now wraps a reference that mentions a dimension name *free* in an `aggregate` whose `output_idx` are the dimension names in shape order; a reference that mentions none — a literal, a parameter expression, an explicit gather, or a gather that *rebinds* the dimension name as its own loop symbol — is evaluated exactly as before. The spec now says what the name is bound to: the 1-based position along that axis, the index space `coords` reads under convention 1, so `index(table, lev)` and `sin(π (x − ½)/N)` work with no gather.

2. **Override keys resolve to the longest dotted suffix (esm-spec §6.6.2 rule 2), and a self-qualified subsystem reference resolves in the Rust array build (§4.6).** Rule 2 tried only the key's trailing segment, so the fully-qualified spelling of a mounted subsystem parameter, `P.wrf.g`, was "Invalid parameter" in Rust (whose single-model array build carries it as `wrf.g`) while Python and Julia (always flattened to `P.wrf.g`) took it as an exact hit. Rule 2 now tries every dotted suffix most-qualified first in all three bindings; rule 3 stays bare-only so `Missing.solo` remains unknown. The same divergence had a second face: `P.wrf.g` inside an *equation* of a single-model document was unbound in Rust's array build; Rust now brings a self-qualified reference (`<model>.<local>` where `<local>` is a declared variable or is rooted at a declared subsystem) back to its local spelling before the build.

## Spec and conformance

- `esm-spec.md` §6.6.2 rule 2 and §6.6.5 `reference` paragraph; `CONFORMANCE_SPEC.md` §5.25 / §5.26; `ESM_COMPLIANCE_VALIDATION_MATRIX.md` BEHAV-06-B-009 / -010.
- New category `tests/conformance/pde_inline_reference_dimension_names/` (Julia-minted golden; one exact decay field asserted through the free-name analytic form at t=0 and t=1, a `const` table lookup by the dimension name, the explicit gather, a gather that rebinds the name, and a reference-free `mean`). `assertion_atol` is 1e-9 rather than the sibling categories' 1e-11 because assertion 2's actual *is* the integrator residual at t=1 (Tsit5 1.8e-14, RK45 5.2e-11); the manifest says so.
- `tests/conformance/override_key_diagnostics/` gains the resolved case `Doc.Left.solo` (binds `Left.solo`).
- Per-binding runners for both categories in Julia, Python, and Rust; unit tests for `bind_dimension_names` (wrap-only-free, capture-aware) and for every override spelling (`P.sub.g`, `sub.g`, `g`) in all three bindings.

## Verification

- Rust: `cargo test --release` in `pkg/earthsci-ast-rs` — 123 test targets, 1423 passed, 0 failed (includes the new `pde_inline_reference_dimension_names_conformance` runner and the extended `override_key_diagnostics_conformance`).
- Python: `pytest tests/test_pde_inline_tests.py tests/test_override_key_suffix_rule.py tests/test_pde_inline_reference_dimension_names_conformance.py tests/test_override_key_diagnostics_conformance.py tests/test_pde_inline_ic_param_override_conformance.py` — 42 passed.
- Julia: `test/conformance_pde_inline_reference_dimension_names_test.jl` 52/52 (reproduces the golden it minted), `test/pde_inline_tests_test.jl` all testsets pass (incl. the new "§6.6.5 reference binds the field's dimension names" 11/11), `test/conformance_override_key_diagnostics_test.jl` 31/31 with the new `Doc.Left.solo` case.
- Downstream: with this branch the EqWeFiC YSU column component's inline references can be written as `index(table, lev)` and its subsystem constants overridden as `P.wrf.g` in every binding (the workaround spellings keep working).

Independent of #177 (shaped state-dependent observeds in assertions), which the same spike needed; the two branches touch different code and either can land first. The remaining gap from the spike — array-valued `default` / `parameter_overrides` / `initial_conditions` — is issue #178.

https://claude.ai/code/session_01XSapCkofqnYF33S5dmFFEH


ctessum:
Closing as superseded by #202, which carries this PR's commit `29dd7bc90` plus four review fixes on top of it.

Containment was verified file-by-file, not assumed:

- `git merge-base --is-ancestor 29dd7bc90 origin/claude/pr-179-review-ube1k3` → true, and the merge-base of the two branches **is** `29dd7bc90`.
- All 21 files were checked individually for later reversion. 11 are byte-identical in #202, including every load-bearing new artifact: the whole `tests/conformance/pde_inline_reference_dimension_names/` category, all three conformance runners, and the `Doc.Left.solo` case in `override_key_diagnostics`.
- The 10 files that differ are modifications, not reversions — two are `cargo fmt` only; the rest are #202's fixes plus renumbering (`BEHAV-06-B-009/-010` → `-012/-013`, `§5.25/§5.26` → `§5.30/§5.31`) because `main` claimed those identifiers in the meantime.

Nothing here is lost by closing.

For the record, this PR's red CI was entirely stale-base and fork artifacts, none of which #202 inherits: the lint failure was an unused `pathlib.Path` import in a file not in this commit; `rust-tests (stable)` failed at the **WASM build** on a `#[cfg(not(target_arch = "wasm32"))]` field with an ungated use; `rust-tests (beta/1.89.0)` were five missing-struct-field errors in `performance.rs` from fields `main` added after this branched. The `Package Security Scan` and `conformance-results` failures were partly `403 Resource not accessible by integration` — fork PRs get a read-only `GITHUB_TOKEN`, so those can never pass from this branch regardless of code.

Follow-on work found while reviewing this line and filed separately: #226 (Julia const-array name clash) and #227 (`sub.g` resolves in Rust but raises in Python/Julia).

PR #180, opened 2026-09-05
ci(security-scan): improve PR reporting and error handling

ctessum:
## Summary
Improves the security scan workflow's reliability and PR integration by fixing error handling in scanner scripts and enabling proper PR comment posting.

## Key Changes

- **Permissions**: Added `pull-requests: write` permission to allow posting security reports as PR comments (previously only `issues: write` was granted, which doesn't cover PR comments)

- **Error handling**: Fixed exit code capture in both security scanner and signature verifier steps to prevent `bash -e` from aborting before outputs are written. Exit codes are now captured via `|| CAPTURE_VAR=$?` pattern instead of after the command completes.

- **PR reporting robustness**: Added `continue-on-error: true` to the "Post security report as comment (PR)" step with explanatory comment. This ensures that read-only tokens on fork PRs don't mask the scan's own verdict.

- **License allowlist**: Added `AGPL-3.0-only` and `AGPL-3.0-or-later` to the security configuration's allowed licenses.

- **Documentation**: Updated workflow comments to clarify that PR comment posting requires `pull-requests:write` permission, not just `issues:write`.

- **File formatting**: Fixed trailing newline in `.security-config.json`.

## Implementation Details

The error handling improvements ensure that scanner exit codes are properly captured before the step completes, allowing subsequent conditional logic to correctly determine success/failure status and write appropriate outputs to `$GITHUB_OUTPUT`.

https://claude.ai/code/session_01J5XupmGHrATh5cZ9pAoBTC

### DIFF
--- a/.github/workflows/security-scan.yml
+++ b/.github/workflows/security-scan.yml
@@ -32,11 +32,14 @@ on:
         default: false
         type: boolean
 
-# The scan job opens an issue when it finds a high-severity vulnerability,
-# so it needs issues:write -- but nothing here needs write access to code.
+# The scan job opens an issue when it finds a high-severity vulnerability and
+# posts its report on the PR -- an issue comment on a PR is gated on
+# pull-requests:write, not issues:write -- but nothing here needs write access
+# to code.
 permissions:
   contents: read
   issues: write
+  pull-requests: write
 
 concurrency:
   group: security-scan-${{ github.ref }}
@@ -125,12 +128,14 @@ jobs:
           SCANNER_ARGS="$SCANNER_ARGS --fail-on-warnings"
         fi
 
+        # `bash -e` would abort the step the moment the scanner exits
+        # non-zero, before the outputs below are written -- so capture the
+        # status instead of letting it propagate here.
+        SCAN_EXIT_CODE=0
         python3 scripts/package-security-scanner.py $SCANNER_ARGS \
-          --output security-scan-results.json
+          --output security-scan-results.json || SCAN_EXIT_CODE=$?
 
         # Extract results for GitHub output
-        SCAN_EXIT_CODE=$?
-
         if [ $SCAN_EXIT_CODE -eq 0 ]; then
           echo "status=success" >> $GITHUB_OUTPUT
         else
@@ -160,11 +165,10 @@ jobs:
       run: |
         echo "Running package signature verification..."
 
+        SIGNATURE_EXIT_CODE=0
         python3 scripts/package-signature-verifier.py \
           --config $SECURITY_CONFIG \
-          --output signature-verification-results.json
-
-        SIGNATURE_EXIT_CODE=$?
+          --output signature-verification-results.json || SIGNATURE_EXIT_CODE=$?
 
         if [ $SIGNATURE_EXIT_CODE -eq 0 ]; then
           echo "Signature verification completed successfully"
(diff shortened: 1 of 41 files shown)


ctessum:
## CI status on this branch: what is this PR's, and what is not

**What this PR set out to do is done and green.** The `Package Security Scan` check now reports 4 packages, 4 passed, 0 warnings, 0 failed (see the scan's own comment above on `f06f235`), against 0 passed / 3 warnings / 1 failed before it.

Everything else red on this branch is red on `main` too, at the same commits, with byte-identical errors. Recording it here rather than fixing it, because none of it is in this PR's diff — this PR touches no Rust, no Go and no TypeScript source.

### `rust-tests` (stable, beta, 1.89.0) and `conformance-results`

`cargo build --target wasm32-unknown-unknown --features wasm` does not compile:

```
    --> src/problem.rs:1179:28
error[E0282]: type annotations needed
    --> src/problem.rs:1182:17
```

`ProblemOptions::const_arrays` is declared `#[cfg(not(target_arch = "wasm32"))]` (`src/problem.rs:253`), but the F32 precision-rounding loop that reads it carries no such gate. The neighbouring reads are gated correctly — `wants_build_pipeline` at `src/problem.rs:1481` has the attribute — so this is one missed gate, introduced by 0bbe895 on 2026-09-01. The second error is a consequence of the first.

`main`'s own `rust-tests (stable)` fails with the identical error, so this is not something this branch introduced. `conformance-results` only aggregates those jobs and goes red with them.

**It is not the one-line fix it looks like.** I reproduced it locally against `wasm32-unknown-unknown` with CI's exact flags, added the missing `#[cfg(not(target_arch = "wasm32"))]` to that loop, and the lib then compiles — but the `esm` binary behind it fails with **11 further wasm32 errors** (E0282, E0425, E0433, E0560, E0609) that the lib failure was masking. Getting the WASM target building again is a real piece of work on the Rust crate, not a patch to land in a security PR. No fix for it exists yet that I could port here.

`rust-tests (beta)` and `rust-tests (1.89.0)` fail at `Run Rust tests`, a different step from the WASM one; those are also red on `main` and I have not diagnosed them.

### `Lint and format (all languages)`

This job is a chain of gates in one `bash -e` step, and each failing gate has been hiding the next since 2026-09-01. Measured locally with the versions CI pins (ruff 0.16.4, Rust 1.98.0, node 20):

| # | Gate | State |
|---|------|-------|
| 1 | `ruff check` | **fixed here** (f06f235) — one unused import from bcffb7e |
| 2 | `ruff format --check` | **fixed here** (8e07dfc) — 6 files, verified AST-identical |
| 3 | `cargo fmt --all -- --check` | ❌ **27 files, 67 hunks** |
| 4 | `cargo clippy --all-targets -- -D warnings` | ❌ 2 errors (`map` over `inspect`; a 9/7-argument fn) |
| 5 | `npm run lint` | ❌ 1 error (unused `name`, `src/data-source-urls.test.ts:71`) + 154 warnings |
| 6 | `npm run format:check` | ❌ 1 file (`src/join-syms.test.ts`) |
| 7 | `gofmt` | ✅ clean |
| 8 | `golangci-lint` | not measured |
| 9 | Version sync | ✅ clean — all six sites agree on 0.2.0 |

Gates 3–6 have not executed in CI since 2026-09-01, so the drift behind them accumulated unobserved. The workflow already carries a comment about exactly this failure mode at the clippy step ("the clippy step sat behind the rustfmt gate, so it has not actually run in…"), which is the same trap one layer down.

Worth noting the shape is the same defect this PR fixes in `security-scan.yml`: several commands sharing one `bash -e` step, where the first failure hides everything after it. Splitting this job's gates into separate steps — or running each with its own `continue-on-error` and a final aggregation — would make the whole set visible in one run instead of one per CI cycle.

---

Issue #172, opened 2026-08-25
operator_compose step-4 conformance (f6860558c) breaks the ReSEACT build: E_TREEWALK_DUPLICATE_DERIVATIVE


ctessum-claude:
## Symptom

Building `reseact.esm` (any grid — reproduced at 6x6x8 and 13x7x72, on ccc0232/ccc0233/ccc0236) with EarthSciAST at current `main` fails in `build_evaluator(form=:oop)`:

```
ERROR: LoadError: E_TREEWALK_DUPLICATE_DERIVATIVE: SuperFast.O3[1,1,1]
Stacktrace:
 [1] _mark_box_covered! at src/tree_walk/stencil_affine.jl:886
 [2] _try_affine_stencil at src/tree_walk/stencil_affine.jl:995
 [3] _compile_arrayop_equation! at src/tree_walk/build.jl:3687
 [5] _compile_derivative_equations at src/tree_walk/build.jl:3172
```

## Bisect (automated, 8 rounds over dc555132c..main, test = full ReSEACT 6x6x8 build)

First bad commit: **f6860558c** "fix(julia): conform operator_compose to the amended step 4, and close the exemptions the fixture fixes resolved" (coupling_apply.jl +316/−, flatten.jl, namespacing.jl). Its parent builds ReSEACT cleanly; every later commit fails.

ReSEACT lifts the SuperFast mechanism onto the grid with `operator_compose` + pointwise lifting; after the amended-step-4 change the tree walk sees two derivative claims for the composed species states (`SuperFast.O3` is simply the first one it visits).

The raising site (`_mark_box_covered!`) is only the detector — the duplicate coverage is created upstream in the composed equation set.

## Repro

```bash
cd code/reseact.esm
export RESEACT_RXENV=<env dev-linking EarthSciAST at main>   # NOT run-model-jl
RESEACT_NLON=6 RESEACT_NLAT=6 RESEACT_NLEV=8 RESEACT_ADJ_STAGES=none \
julia --project=$RESEACT_RXENV -e '
  const REPO = pwd()
  include(joinpath(REPO, "tools", "diag", "_env.jl"))
  Base.include(Core.eval(Main, :(module _Drv end)), joinpath(REPO, "tools", "adjoint_gradient.jl"))'
```

Pinning EarthSciAST to f6860558c^ (or any pre-phase-6-step-4 commit) builds and runs correctly.

[...]

PR #182, opened 2026-09-05
fix(flatten): attribute a merged operator_compose equation to its variable's system

### LINKED ISSUE #173: Chained operator_compose with the operator as A does not merge in the Python oracle (and TS; corpus cannot pin it)
## Symptom

Chained `operator_compose` entries over one state, with the OPERATOR as system A, do not fully merge in the Python oracle (and therefore cannot be pinned by the shared flatten corpus). Julia handles the pattern as of the #172 fix; the other bindings are unaudited for it.

The pattern is not hypothetical — it is exactly how `reseact.esm` composes SuperFast:

```json
"coupling": [
  {"type": "operator_compose", "systems": ["SuperFastDepositionSink", "SuperFast"]},
  {"type": "operator_compose", "systems": ["SuperFastEmissionSource", "SuperFast"]},
  {"type": "operator_compose", "systems": ["SuperFast", "Transport3D"], "lifting": "pointwise"}
]
```

with the sink/source components authoring `D(SuperFast.X)` equations directly (B's variable, A's authorship — the provenance case `advection_reaction_loaded_ic_bc.esm` pins in the other direction).

## Mechanism (Python)

`_apply_operator_compose` (flatten.py) merges into `a.equations[i]` — the merged equation stays in **A's component bucket**. The next compose entry walks `components[b_name].equations` and `components[a_name].equations` only, so the composed tendency filed under the previous entry's operator is invisible to it. Both the un-merged contribution and the composed equation survive to `_assemble_system`, which raises:

```
ConflictingDerivativeError: Two systems define non-additive equations for variable 'Chem.O3': Sink vs Src
```

Loud rather than corrupt — but the document is spec-valid and should flatten.

## Repro

Minimal 4-system document (Chem with two species, Sink and Src authoring `D(Chem.*)` directly, Adv with a `_var` template; chained `[Sink,Chem]`, `[Src,Chem]`, `[Chem,Adv]` pointwise): the Julia test `flatten_conformance_test.jl` testset "operator_compose: CHAINED entries over one state keep merging" (added with the #172 fix) carries it verbatim, with the expected flattened answer:

```
D(Chem.NO)/Dt = (-Sink.kd) * Chem.NO + (-Chem.k2) * Chem.NO + (-Adv.wind) * grad(Chem.NO)
D(Chem.O3)/Dt = Src.e + (-Sink.kd) * Chem.O3 + (-Chem.k1) * Chem.O3 + (-Adv.wind) * grad(Chem.O3)
```

## What Julia does (the model for the fix)

After summing, the merged equation is reattributed to the system that OWNS its dependent variable (`_component_root(dep)`), on the reasoning that a composed tendency is no longer any single author's contribution. Where A already is the variable's system — every current corpus case — the reattribution is the identity, so nothing else changes. The Python analogue is relocating the merged equation to the variable owner's component bucket (position semantics need care: §4.7.5 step 4 makes document order normative).

## Scope

- [ ] Python oracle: relocate the merged equation so chained entries can see it
- [ ] TypeScript: same component-bucket architecture as Python (`applyOperatorCompose` merges into `components[A].equations`, flatten.ts:1259) — shares the gap structurally; needs the same relocation
- [ ] Rust / Go: audit for the same gap (all post-efcc35594 implementations are ownership-aware and likely share it)
- [ ] Corpus: add a chained-compose tier case once the oracle agrees, so the Julia-local pin can be promoted to a cross-language one
- [ ] Spec: §4.7.1 says nothing about the ownership of a merged equation; the amendment should state the reattribution rule

Found while fixing #172 (the Julia tree-walk duplicate-derivative regression); Julia-side behavior is pinned locally in `flatten_conformance_test.jl` until the corpus can carry the case.

ctessum:
Fixes #173.

## The gap

Chained `operator_compose` entries over one state, with the OPERATOR as system A, stopped merging after the first entry in the Python oracle, TypeScript, Rust and Go. The shape is not hypothetical — it is how `reseact.esm` composes SuperFast:

```json
"coupling": [
  {"type": "operator_compose", "systems": ["SuperFastDepositionSink", "SuperFast"]},
  {"type": "operator_compose", "systems": ["SuperFastEmissionSource", "SuperFast"]},
  {"type": "operator_compose", "systems": ["SuperFast", "Transport3D"], "lifting": "pointwise"}
]
```

with the sink and source authoring `D(SuperFast.X, t)` directly — B's variable under A's authorship, which §4.7.1 permits.

All four bindings merge into A's equation, and in all four *"who authored this equation"* **is** which component's equation list it sits in. So entry 1's sum stayed filed under the sink; entry 2 walked the source's and the chemistry's lists, could not see it, and merged nothing. Both the composed tendency and the source's un-merged contribution reached assembly:

```
ConflictingDerivativeError: Two systems define non-additive equations for variable 'Chem.O3': Sink vs Src
```

Loud rather than corrupt — but the document is spec-valid and must flatten.

## The fix

Julia already does the right thing (the #172 fix): after summing, it reattributes the merged equation to the system that owns its dependent variable, on the reasoning that a composed tendency is no longer any single author's contribution. The other four now do the same, by relocating the equation into that component's list.

It is **appended** to the owner's list, which is what reproduces Julia's order: each merge trails the ones before it, so a species whose chain ends at a later entry lands after one whose chain ended earlier. Inserting at the consumed B equation's slot instead would freeze both at the state's original declaration order and disagree with every other binding.

Where A already is the variable's system — every corpus case before this one — the move is the identity. Regenerating the corpus confirms it: the only change to `cases.json` is the new case (+124 lines, 0 modified).

## What's in the change

| | |
|---|---|
| `esm-libraries-spec.md` | §4.7.1 step 4 gains the attribution rule. It said nothing about the ownership of a merged equation, which is why four bindings could disagree with the reference and still read as correct. |
| `tests/coupling/chained_operator_compose.esm` | The shared fixture — the minimal 4-system document from the issue. |
| `tests/conformance/flatten/cases.json` | Carries it as the new `chained_compose` tier, so the pin that has been Julia-local since #172 is now cross-language. |
| `pkg/EarthSciAST.jl/test/flatten_conformance_test.jl` | That testset now drives the shared file instead of an inline copy. |
| `flatten.py` / `flatten.ts` / `flatten.rs` / `flatten.go` | `_reattribute_merged_equations` and its three counterparts, called after the merge loop (positions are collected during it, because A's index is live until the last B equation is matched). Self-compose (`["X", "X"]`, which nothing rejects) is skipped: the shared list has just been rebound, and the move would be the identity there anyway. |
| `pkg/earthsci-ast-py/tests/test_flatten.py` | A fixture-independent regression test. The corpus is *generated* from the Python binding, so a regression there would rewrite the expected answer rather than fail against it. |

## Verification

The flattened answer now agrees with Julia's, exactly:

```
D(Chem.NO)/Dt = (-Sink.kd) * Chem.NO + (-Chem.k2) * Chem.NO + (-Adv.wind) * grad(Chem.NO)
D(Chem.O3)/Dt = Src.e + (-Sink.kd) * Chem.O3 + (-Chem.k1) * Chem.O3 + (-Adv.wind) * grad(Chem.O3)
```

- **Python** — full suite green (3133 passed, 23 skipped, 3 xfailed); `ruff check` / `ruff format --check` clean on every touched file.
- **TypeScript** — full suite green (2539 tests); prettier clean.
- **Go** — `go test ./...` green; `gofmt` clean.
- **Rust** — `cargo test --test flatten_conformance` green (5/5); `cargo fmt --check` clean for `flatten.rs`.

Each of Go and TypeScript was also re-run with its fix reverted, and both fail on the new corpus case — so the case is doing its job rather than passing vacuously.

Julia is unchanged by this PR (it has been correct since #172); its conformance run in CI is what confirms the new corpus case against the reference implementation.

## Notes for review

- Equation `source_system` is deliberately left as A's on a merge: Python, TypeScript and Go all already did that, the corpus compares the field for those three, and it is the provenance the `ConflictingDerivativeError` diagnostic reads. The *attribution* that moved is the component bag, which is the analogue of Julia's per-equation `owners` vector.
- The two attributions genuinely coincide in the bucket-based bindings, so relocating changes position as well as ownership. The spec text says so explicitly rather than claiming position is untouched.

https://claude.ai/code/session_01Cy5mPaV6B54dPUP5v7eYt2

---
_Generated by [Claude Code](https://claude.ai/code/session_01Cy5mPaV6B54dPUP5v7eYt2)_

### DIFF
--- a/esm-libraries-spec.md
+++ b/esm-libraries-spec.md
@@ -768,6 +768,40 @@ order; the application sequence and the reporting sequence are separate.
      already substituted, so in both cases this rewrite is the identity. Step 4 introduces no
      renaming of its own beyond it.
    - Variables from system B that appear in the combined RHS are added to the merged system's variable list.
+   - **The merged equation belongs to its dependent variable's system.** A composed tendency is
+     no longer either author's contribution, so for the purpose of every LATER `operator_compose`
+     entry the merged equation is attributed to the system that owns `x` — the component named by
+     `x`'s leading namespace segment — and not to whichever of A and B happened to carry the
+     surviving equation.
+
+     Where A already is that system, this is the identity and there is nothing to observe. It bites
+     where A is not: §4.7.1 places no restriction on which side is the operator, and an operator
+     component may author `D(Chemistry.O3, t) = …` directly — B's variable under A's authorship.
+     A document may then CHAIN entries over one state with the operator as `systems[0]`:
+
+     ```json
+     {"type": "operator_compose", "systems": ["DepositionSink",  "Chemistry"]},
+     {"type": "operator_compose", "systems": ["EmissionSource",  "Chemistry"]},
+     {"type": "operator_compose", "systems": ["Chemistry", "Transport"], "lifting": "pointwise"}
+     ```
+
+     Entry 1 consumes `Chemistry`'s own equation for `O3`. If the sum is left attributed to
+     `DepositionSink`, entry 2 looks at `EmissionSource` and `Chemistry`, finds no equation for
+     `O3` in either, and merges nothing — so the emission term is never added and the flattened
+     system carries TWO equations claiming `D(Chemistry.O3, t)`. That is an over-determination
+     raised (at best) stages later, for a document that is spec-valid and must flatten.
+
+     Attribution is not recoverable from the equation itself: after namespacing, an operator's
+     equation for a chemistry species is textually indistinguishable from the chemistry
+     component's own, and placeholder expansion produces the same shape. An implementation MUST
+     therefore carry it explicitly — as a per-equation owner beside the equation list, or as which
+     component's equation list the equation lives in — and MUST update it here.
+
+     Attribution is about VISIBILITY to later entries, not about ordering; ordering stays what
+     §4.7.5 step 4 says. Note that in an implementation whose attribution IS the equation list an
+     equation sits in, updating it necessarily moves the equation: append it to the owner's list,
+     so that in a chained composition each merged equation trails the merges that preceded it.
+
    - **The merged-away name does not survive.** When a match renames the dependent variable — a
      translation match, or the bare-name fallback, which is a name-based translation — B's
      declaration of that variable is left with nothing to constrain it: the equation that defined
--- a/pkg/EarthSciAST.jl/test/flatten_conformance_test.jl
+++ b/pkg/EarthSciAST.jl/test/flatten_conformance_test.jl
@@ -861,38 +861,15 @@ end
         # dependent variable, not left as A's authored contribution. Keeping
         # A's ownership made entries 2 and 3 match nothing, and the flattened
         # system carried THREE equations claiming `D(Chem.O3)`.
-        doc = """
-        {"esm": "1.0.0",
-         "metadata": {"name": "ChainedCompose",
-                      "description": "chained operator_compose with the operator as A",
-                      "authors": ["EarthSciAST test suite"],
-                      "created": "2026-08-25T00:00:00Z"},
-         "models": {
-           "Chem": {"variables": {"O3": {"type": "unknown", "units": "mol/mol", "default": 1.0},
-                                  "NO": {"type": "unknown", "units": "mol/mol", "default": 1.0},
-                                  "k1": {"type": "parameter", "units": "1/s", "default": 0.1},
-                                  "k2": {"type": "parameter", "units": "1/s", "default": 0.2}},
-                    "equations": [{"lhs": {"op": "D", "args": ["O3"], "wrt": "t"},
-                                   "rhs": {"op": "*", "args": [{"op": "-", "args": ["k1"]}, "O3"]}},
-                                  {"lhs": {"op": "D", "args": ["NO"], "wrt": "t"},
-                                   "rhs": {"op": "*", "args": [{"op": "-", "args": ["k2"]}, "NO"]}}]},
-           "Sink": {"variables": {"kd": {"type": "parameter", "units": "1/s", "default": 0.01}},
-                    "equations": [{"lhs": {"op": "D", "args": ["Chem.O3"], "wrt": "t"},
-                                   "rhs": {"op": "*", "args": [{"op": "-", "args": ["kd"]}, "Chem.O3"]}},
-                                  {"lhs": {"op": "D", "args": ["Chem.NO"], "wrt": "t"},
-                                   "rhs": {"op": "*", "args": [{"op": "-", "args": ["kd"]}, "Chem.NO"]}}]},
-           "Src":  {"variables": {"e": {"type": "parameter", "units": "mol/mol/s", "default": 2.0}},
-                    "equations": [{"lhs": {"op": "D", "args": ["Chem.O3"], "wrt": "t"},
-                                   "rhs": "e"}]},
-           "Adv":  {"variables": {"wind": {"type": "parameter", "units": "m/s", "default": 3.0}},
-                    "equations": [{"lhs": {"op": "D", "args": ["_var"], "wrt": "t"},
-                                   "rhs": {"op": "*", "args": [{"op": "-", "args": ["wind"]},
-                                                               {"op": "grad", "args": ["_var"]}]}}]}},
-         "coupling": [{"type": "operator_compose", "systems": ["Sink", "Chem"]},
-                      {"type": "operator_compose", "systems": ["Src", "Chem"]},
-                      {"type": "operator_compose", "systems": ["Chem", "Adv"], "lifting": "pointwise"}]}
-        """
-        flat = flatten(load_string(doc))
+        #
+        # The document moved into the shared tree with the issue #173 fix (the
+        # same gap in the Python oracle, TypeScript, Rust and Go), so the corpus
+        # now carries it as the `chained_compose` tier and this testset drives
+        # the SAME file. What stays here is the rendered answer: the corpus
+        # records it too, but reading it as prose is what makes a regression
+        # legible.
+        flat = flatten(load_path(joinpath(_FLATTEN_TESTS_DIR, "coupling",
+                                          "chained_operator_compose.esm")))
         # ONE equation per species, every chained contribution in its RHS. The
         # POSITIONS are the merge trail: each merge keeps the surviving A
         # equation's document slot, so NO's chain ends at the sink's slot and
--- a/pkg/earthsci-ast-go/pkg/esm/flatten.go
+++ b/pkg/earthsci-ast-go/pkg/esm/flatten.go
@@ -1436,6 +1436,12 @@ func applyOperatorCompose(components map[string]*componentSystem, entry Operator
 	// bDep -> targetDep for every match that RENAMED the dependent variable.
 	mergedAway := map[string]string{}
 	var mergedAwayOrder []string
+	// Positions in A's equation slice this entry merged INTO, for the
+	// reattribution below. Collected rather than acted on in place because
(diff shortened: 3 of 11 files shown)


ctessum:
### `Package Security Scan` is failing, and it is not this PR's

Standing down on it rather than fixing it here, with the evidence.

**It is red on the base branch.** Commit `3a01c56` (current `main`) produces a byte-identical scan summary to this PR's run — 4 packages, 0 passed, 3 warnings, 1 failed — and exits 1 at the same step:

```
❌ earthsci-ast-ts: FAILED
   🔴 1 vulnerabilities found
   - fast-uri: unknown (high)
```

The three most recent `main` runs ([903](https://github.com/EarthSciML/EarthSciAST/actions/runs/33982406830), [902](https://github.com/EarthSciML/EarthSciAST/actions/runs/33981896596), [901](https://github.com/EarthSciML/EarthSciAST/actions/runs/33974661412)) are all `failure`, and it has been failing on `main` on and off since run 893.

**This PR cannot have caused it.** `git diff origin/main...HEAD` over every manifest and lockfile in the repo — `package.json`, `package-lock.json`, `Cargo.toml`, `Cargo.lock`, `pyproject.toml`, `uv.lock`, `go.mod`, `go.sum`, `Project.toml`, `Manifest.toml` — is empty. The diff adds no dependency to anything.

**No fix exists to port**, so there is nothing for me to pull in. Fixing it means a lockfile change in `pkg/earthsci-ast-ts`: `fast-uri@3.1.5` is transitive, pulled by `ajv@8.20.0` (`^3.0.1`), so `npm audit fix` (or a targeted `npm update fast-uri`) in that package is the patch. That is a dependency bump unrelated to this change, and doing it here would widen the PR into territory a flatten fix has no business in — it belongs on `main` where the failure actually lives.

**A second, separate defect shows only on the PR trigger.** The job's *final* error on this PR is not the scan at all — it is the report-comment step 403ing:

```
RequestError [HttpError]: Resource not accessible by integration
  POST /repos/EarthSciML/EarthSciAST/issues/182/comments   status: 403
  x-accepted-github-permissions: 'issues=write; pull_requests=write'
```

`security-scan.yml` already declares `permissions: issues: write` (line 39), so the workflow file is not the problem — the token is being capped below what it asks for, which points at the repository's Actions → Workflow permissions setting being read-only. This step never runs on a `push` to `main`, which is why the failure looks different there. Worth fixing separately, either by lifting that setting or by marking the comment step `continue-on-error: true` so a reporting failure cannot mask a scan result.

**I have not spent a re-run on it.** The failure is a deterministic dependency-advisory lookup plus a deterministic permissions 403 — neither is a flake, and reproducing it on `main` is the stronger form of the same check. Say the word if you would like it re-run anyway.

Every other check on this PR is expected to be green; the substantive verification is in the PR description.

---

ctessum:
### The Rust and Lint checks are also red on `main`, for three unrelated reasons

Following on from the security-scan note. `rust-tests (stable)` failed here, so I traced it — and the same [`main` run at `3a01c56`](https://github.com/EarthSciML/EarthSciAST/actions/runs/33982406818) shows `rust-tests` red on **all three** toolchains plus `Lint and format`, at the same steps. None of it is reachable from this diff, which touches one function in `flatten.rs` and no manifest, lockfile, or shared type.

**One operational consequence worth flagging first:** between these two failures, **CI is not currently running the Rust test suite at all** — not on this PR and not on `main`. On `stable` the WASM build fails first, so `Run Rust tests` is skipped; on `beta` and `1.89.0` `cargo test` aborts while compiling a test target, so no test binary ever runs. My Rust change is therefore covered only by the local run reported in the PR description (667 lib + 5 conformance tests green).

---

**1. `rust-tests (stable)` — WASM build, `problem.rs:1179`**

```
    --> src/problem.rs:1179:28
     = note: available fields are: `p`, `u0`, `model_name`, `metaparameters`, `base_path` ... and 4 others
```

`ProblemOptions::const_arrays` is declared `#[cfg(not(target_arch = "wasm32"))]` (`problem.rs:252-253`), but the rounding loop that reads it is not gated, so it cannot compile for `wasm32`. `git log -L 1174,1186:src/problem.rs` puts that block at `0bbe895` ("wip(rust): per-variable element_type"). Proposed patch — gate the statement to match the field:

```rust
+       #[cfg(not(target_arch = "wasm32"))]
        for (k, a) in opts.const_arrays.iter_mut() {
```

**2. `rust-tests (beta)` and `(1.89.0)` — `tests/performance.rs` stale against the type changes**

Five `E0063`s: `Species` is missing `default_units` (lines 526, 535, 544), `Parameter` is missing `default_units`, `distribution`, `shape` and one more (line 577), and `ReactionSystem` is missing `analyses`, `tests`, `tolerance` (line 584). The struct-literal test fixtures were not updated when those fields were added. All three types derive `Default` (`src/types/components.rs:7, 59, 94`), so the minimal fix is `..Default::default()` on each of the five initializers — which also stops this test target rotting the next time a field is added.

**3. `Lint and format (all languages)` — `ruff check`**

Fails at the very first lint step, which is why `cargo fmt`, `cargo clippy`, eslint/prettier, gofmt and golangci-lint are all reported as *skipped* rather than passing. I reproduced it locally with the workflow's own invocation:

```
F401 `pathlib.Path` imported but unused
  --> tests/test_conformance_round_trip.py:25:21
```

plus `ruff format --check` drift in `src/earthsci_ast/numpy_interpreter.py`, `src/earthsci_ast/structural_checks.py`, `tests/test_aggregate_conformance.py`, `tests/test_closed_functions.py`, `tests/test_data_source_url_conformance.py` and `scripts/run-property-corpus-conformance.py`. `ruff check --fix` and `ruff format` on those paths clears it. Every file this PR touches is already clean under both.

---

I am not pushing any of these. Each is on `main`, each is in code this change does not touch, and folding a wasm cfg gate, a test-fixture refresh and a repo-wide ruff pass into a flatten fix would make the diff unreviewable for its actual subject. Happy to open them as a separate PR if you want — the fixes above are small and I have verified the facts each rests on. Say the word and I will.

---

ctessum:
### `conformance-results` — downstream of the Rust failures, plus that 403 again

Last of the red checks, and it is a consequence of the ones above rather than a new problem. The chain:

1. `standard-conformance-testing` declares `needs: [julia-tests, typescript-tests, python-tests, rust-tests, go-tests]` with no `if: always()`, so the failing `rust-tests` legs **skip** it.
2. `conformance-results` is `if: always()`, so it runs, finds no artifact to download, and — on a `pull_request` — dies at *Comment PR with enhanced conformance results* with the same `403 Resource not accessible by integration` that kills `Package Security Scan`. Its `Check final conformance status` step never even ran (skipped).

So on this PR the job's only actual failure is the permissions 403. On `main` the same job fails one step later, at the status check, because the comment step doesn't run for a `push`. Same root, different door — and the 403 is now confirmed across two separate workflows, which is why it looks like the repo-level Actions token setting rather than anything in either workflow file.

**One consequence worth separating out**, because it sounds worse than it is: with `standard-conformance-testing` skipped, `scripts/test-conformance.sh` — the *cross-binding comparison* harness — has not run in CI on this PR or on `main`. That does **not** mean the new corpus case is unverified: each binding drives `tests/conformance/flatten/cases.json` from inside its own suite, and those all ran and passed here — `julia-tests` on 1.10, 1.11 and 1.12, `python-tests` on 3.9–3.12, `typescript-tests` on 20 and 22, `go-tests` on 1.21–1.23. Four of the five bindings confirm the `chained_compose` case in CI. Only Rust's is missing, for the reason above, and it is covered by the local run in the PR description.

Getting `rust-tests` green would restore the harness for everything, which is a further argument for fixing those two small things on `main`.

That accounts for every red check on this PR. Nothing else outstanding from my side.

---

PR #183, opened 2026-09-05
fix(bindings): §9.7.7 rename must rewrite an `integral` rule's `var` and bare-axis-name bounds

### LINKED ISSUE #174: esm-spec §9.7.7: import renaming does not rewrite an `integral`'s `var`/bounds, so an integral rewrite rule cannot be renamed onto another axis

ctessum:
Fixes #174.

## The bug

§9.7.7's normative occurrence list for an **index-set** rename names only the `wrt`/`dim` axis scalar fields of Expression nodes, and all five bindings implement exactly that list. It omits the axis-naming scalar fields of `integral`: the integration variable `var`, and a `lower`/`upper` bound written as a bare axis name (the §4.2 cumulative form `"upper": "x"`).

The asymmetry the issue reports, reproduced here before the fix (Python binding, a library imported under `rename: {"x": "lev"}`):

```
match after rename:  {"op":"integral","args":["f"],"var":"x","lower":0,"upper":"x"}
where after rename:  {"f":{"shape":["lev"]}}          <- renamed
ranges after rename: {"i":{"from":"lev"}, ...}        <- renamed
```

The shape constraint and the body's `{"from": …}` ranges follow the rename, and a `D` rule's `wrt` would — but `var`/`upper` do not. The renamed instance therefore never matches a consumer that only ever named `lev` (the only vocabulary that consumer knows), nothing fires, and the `integral` survives lowering as `unlowered_operator`. A discretization library that ships an `integral` rule family is unusable through the very mechanism §9.7.7 exists to provide; the only workarounds are copying-and-editing the library file (barred by §9.7.1) or having the consumer write the *library's* axis name in its own equations, which defeats the rename.

## The fix

In all five bindings (Julia, TypeScript, Python, Rust, Go):

- **`var` joins the rename AXIS key set** alongside `wrt`/`dim`, and with it the §9.7.6 metaparameter-substitution skip set — an axis name is not an expression position, so a bound metaparameter of the same name must not rewrite it into an integer (the guard `wrt`/`dim` already carry). This also removes `var` positions from the free-name inventory that backs the `rebind` occurs-check, which is what an axis name should be.
- **A new bound branch** maps a bare `lower`/`upper` string through the index-set map *when it names a renamed index set*. Every other bound — a numeric literal, a parameter reference, an AST subtree — stays an ordinary expression position handled by the metaparameter/rebind substitution, so nothing that isn't an axis occurrence changes.

Julia's canonical `_STRUCTURAL_FIELDS` table gains `"var" => :axis` and `"lower"/"upper" => :bound` (a new kind, documented), so its derived skip/protect sets stay generated rather than hand-maintained.

## Conformance fixture

`tests/conformance/expression_templates/import_rename_integral_axis/` — in the shape of `import_rename_two_instances/` but over an `integral` rule, as the issue suggests. One cumulative-integral rule family (`column_integral_1d.esm`: metaparameter `N`, index set `x`, cell measure `dx`, a `where`-constrained rule matching `{op: "integral", args: ["f"], var: "x", lower: 0, upper: "x"}` that lowers to the §4.3.1 prefix reduction) is imported twice — `prefix: "col"` + `rename: {"x": "lev"}` at `N = 4`, and `prefix: "row"` + `rename: {"x": "lat"}` at `N = 3` — and the consumer writes both integrals in the renamed vocabulary. Each instance fires only on its own axis at its own cell measure (`1/4` on `lev`, `1/3` on `lat`); without the fix neither fires and both integrals survive.

Registered in the golden lists of all five suites and in `scripts/generate-template-import-goldens.jl`, with a targeted per-binding assertion beside the golden comparison.

## Docs

esm-spec §9.7.7 transitivity list (the normative change) and its fixture index; the renaming RFC's fixture list; the fixture README; EXPR-09-F-003 / EXPR-09-F-008 plus a dated note in the compliance matrix.

## Testing

| Binding | Result |
|---|---|
| Python | full suite `3132 passed, 26 skipped, 3 xfailed` |
| Go | full suite `go test ./...` ok |
| TypeScript | full suite `70 files, 2528 tests` passed; `typecheck` + prettier + eslint clean |
| Rust | `--lib` 667 passed; `template_imports_conformance` 28 and `out_of_line_templates_conformance` 12 passed |
| Julia | **not run — no Julia toolchain in this environment** |

Python, Go, TypeScript and Rust each compare the new fixture against the same committed `expanded.esm` and agree on it byte-for-byte, so the golden is pinned four ways; the Julia edits mirror the other four but are unverified here and want CI (or a local `Pkg.test()`) to confirm. The golden was generated with the Python pipeline in the format `scripts/generate-template-import-goldens.jl` writes (sorted keys, 2-space indent, raw UTF-8, no integral-valued float literals per CONFORMANCE_SPEC §canonical bytes); regenerating it under Julia should be a no-op.

## Noted in passing, not changed

Go's `metaSubstSkipKeys` (`pkg/earthsci-ast-go/pkg/esm/metaparams.go`) is missing `"dim"`, which the other four bindings all skip — a pre-existing divergence: a bound metaparameter sharing a name with an axis would be substituted into a `dim` field in Go only. Out of scope here; worth its own fix.

https://claude.ai/code/session_013VXri3JptAhQNzwEbAbc9c

---
_Generated by [Claude Code](https://claude.ai/code/session_013VXri3JptAhQNzwEbAbc9c)_


ctessum:
## CI status: the red checks here are red on `main` too

Recording this so the failures aren't mistaken for fallout from this diff. Every check that has failed so far fails identically on the base branch, at this PR's own base commit `3a01c56` — conformance run [33982406818](https://github.com/EarthSciML/EarthSciAST/actions/runs/33982406818) and security-scan runs 901–903.

| Check | Failure | On `main` @ `3a01c56` |
|---|---|---|
| **Lint and format (all languages)** | `ruff` F401 — `pathlib.Path` imported but unused, `pkg/earthsci-ast-py/tests/test_conformance_round_trip.py:25` | fails at the same `ruff check` step |
| **rust-tests (stable / beta / 1.89.0)** | `tests/performance.rs` fails to compile: `E0063` missing fields `default_units`, `distribution`, `shape` in `Species`/`Parameter`, and `analyses`, `tests`, `tolerance` in `ReactionSystem` | all three fail the same way (stable fails one step earlier, at *Build WASM*) |
| **Package Security Scan** | separate workflow | fails on the last three `main` runs |

None of these touch anything this PR changes: the diff is the §9.7.7 rename walk in the five bindings, one new conformance fixture, their test registrations, and docs. It does not touch `test_conformance_round_trip.py`, `tests/performance.rs`, or the type definitions whose new fields that file is missing. I confirmed the ruff one directly against CI's pinned `ruff==0.16.4`: the import is unused in the committed file, whose last change (`bcffb7e`) is already on `main`, and it is the only error the repo-wide `ruff check src/ tests/ ../../scripts/` reports.

No fix PR exists for any of them to port into this branch, so I'm leaving them alone rather than widening this PR. Each is a small, self-contained fix in its own right — dropping the unused import, and updating `tests/performance.rs` for the fields those structs gained — but they belong to whoever owns that work.

One knock-on worth flagging, because it affects every PR right now: the lint job dies at the `ruff` step, so the `cargo fmt`, `clippy`, `eslint`/`prettier`, `gofmt` and version-sync steps after it are all skipped and currently report nothing on any diff. I ran each of them locally against the files this PR changes and they are clean.

Still pending here, and the checks that actually verify this change: **julia-tests (1.10 / 1.11 / 1.12)**, plus the python / go / typescript matrices (all green on `main`). Julia is the one binding I could not run in my environment, so those jobs are the real verification of the Julia half of this change; python 3.11 has already come back green. I'll follow up if any of them go red.

---

PR #184, opened 2026-09-05
fix(julia): an index gather distributes through NESTED elementwise ops (§4.3.4)

### LINKED ISSUE #175: Julia tree-walk: an elementwise-defined array observed reached only through an aggregate gather fails with E_TREEWALK_UNBOUND_VARIABLE

ctessum:
Fixes #175.

## What this fixes

An ARRAY-shaped **observed** written ELEMENTWISE over another array and read **only** through an `index(f, j)` gather failed to build in Julia:

```
zc : shape [lev]
f  : shape [lev],  f = 1 + cos(pi*zc)
S  : aggregate(j from lev; index(f, j))
```

raised `E_TREEWALK_UNBOUND_VARIABLE: zc`, while Python and Rust returned the right numbers for the same document. An `.esm` library of column geometry that defines its fields elementwise was portable to the other bindings but not to the reference one.

## Where it actually came from

Not `shape_promotion.jl`, which the issue nominated: `f` already carries an array shape, so `promote_downstream_shapes` never touches it and `newly_promoted` is not involved.

The real path is `_fold_elementwise_array_observeds` (`tree_walk/build_helpers.jl`), which inlines an elementwise array observed into its readers **by name substitution**. The reader's gather therefore becomes

```
index(1 + cos(pi*zc), j)
```

`_resolve_indices_op` already carries a push-down branch for exactly this — indexing an elementwise combination IS the combination of the indexed operands, so distributing is exact — but its operand test looked only at the **immediate** operands: `1` is a literal and `cos(pi*zc)` is neither an array producer node nor a variable. Nothing matched, the gather was **dropped**, and the array leaf `zc` reached `_compile` bare. Confirmed by reproducing the error and bisecting it with `git stash`.

That also explains both spellings the issue reports as working: the explicit `aggregate(k from lev; … index(zc, k) …)` gather never needs the push-down, and an elementwise body consumed directly in a state RHS is lifted per-cell by `_index_array_leaves` before this branch is ever reached.

## What is in the change

* **`_index_pushdown_arrayish`** (`tree_walk/build_helpers.jl`) — one operand predicate, shared by `_resolve_indices_op` and the symbolic stencilizer's mirror of the same branch (`_stencilize_index`), which had drifted apart. It recurses through nested elementwise ops, so the gather lands on the leaves as `1 + cos(pi*index(zc, j))`, and it spans every array SOURCE the resolver can gather — an array state slot, a live forcing buffer, and a const array — where the resolver's copy previously knew only about state slots.
* Two correctness refinements fall out of having one predicate rather than two:
  * a **scalar reduction** (an `aggregate` with empty `output_idx`) produces a scalar and is no longer wrapped in a gather, where testing `_is_aggregate_op` alone had wrapped it;
  * a **one-element const-array entry** is a 0-D loader field that a bare reference already const-folds to a literal (RFC pure-io-data-loaders §4.3), so it stays out of the leaf test rather than being gathered out of range past the first cell.
* **New conformance category** `tests/conformance/elementwise_observed_gather/` — fixtures, Julia-minted goldens, and per-binding runners for Julia / Python / Rust, each gating every assertion against **both** the golden actual and the fixture's own declared `expected`. It carries a **controlled pair**: `elementwise_gather.esm` is the shape under test and `explicit_gather.esm` writes the identical field as an explicit `aggregate(k from lev; 1 + cos(pi*index(zc, k)))` — the spelling that never needed the push-down. The two share every assertion and every expected value, and each runner additionally requires them to agree with **each other** actual-for-actual, so a divergence between the two spellings is the push-down and cannot be the physics. Both right-hand sides are state-free, so `u(t) = colsum·t` and `s(t) = total·t` are integrated exactly by every pinned solver family and the goldens are integrator-independent to machine precision.
* **CONFORMANCE_SPEC §5.27** (+ §5.27.1 on what is and is not a gatherable leaf, §5.27.2 the gate) and **ESM_COMPLIANCE_VALIDATION_MATRIX** row BEHAV-04-I-001.
* A Julia unit test (`tree_walk_elementwise_obs_gather_test.jl`) covering the end-to-end fixture, the two working spellings the issue names as differential oracles, and the predicate itself.

## Test results

* **Julia** — full `Pkg.test()` sweep: **22584 passed, 0 failed** (7 pre-existing `@test_broken`), 66m22s. New unit test `17/17`; new conformance adapter `62/62`. Both fail on a `git stash` of the source change, so neither is vacuous. All three CI `julia-tests` legs green (1.10, 1.11, 1.12).
* **Python** — `pytest tests/`: **3135 passed, 25 skipped, 3 xfailed, 0 failed**. New adapter `4 passed`. `ruff check` and `ruff format --check` clean on the new file. All four CI legs green.
* **Rust** — `cargo test --release`: **1421 passed, 0 failed** across 123 test targets — but note this is **default features only**. The three CI `rust-tests` legs test the whole feature surface and never reach the test step, because `tests/performance.rs` does not compile under the `performance` feature and the WASM build fails, both of which are broken on `main` too (see the comments below). So the full Rust feature surface is unverified on this branch by CI — as it is on `main`, and by the same amount. `rustfmt --check` and `cargo clippy` are clean on the new file.

## Pre-existing CI failures

Six checks are red on `main` at this branch's base (`3a01c56`) and are red here for the same reasons; none is caused by this diff, which touches no dependency manifest and no Rust or Python source. Each is root-caused with a proposed patch in the comments below:

* `Lint and format` — a `ruff` F401 in an unrelated test file, which fails on step 1 of 9 and so *skips* `cargo fmt`, `clippy`, `eslint`, `gofmt`, `golangci-lint` and version-sync.
* `rust-tests` × 3 — `performance.rs` E0063 on beta/1.89.0, and a two-layer WASM breakage on stable.
* `Package Security Scan` — a live high-severity `fast-uri` advisory in the TS lockfile, plus a token-permission 403 that stops the scan reporting on PRs at all.
* `conformance-results` — purely downstream: `standard-conformance-testing` is gated on the `rust-tests` legs and was therefore *skipped*, so no results artifact exists, and the job's own "Comment PR" step then hits the same 403 as the security scan. No conformance test actually ran and failed.

Everything else is green: `julia-tests` × 3, `python-tests` × 4, `go-tests` × 3, `typescript-tests` × 2, `generate-docs`.

## Cross-binding note

The category independently confirms the issue's premise rather than taking it on trust: Rust and Python evaluate **both** fixtures correctly and agree with the Julia goldens within tolerance, so this really was a Julia-only gap and Python/Rust are now pinned against regressing into it.

Independent of #177 and #179 — neither touches this code path, and #177's own description lists #175 as filed rather than fixed. Any of the three can land first.

https://claude.ai/code/session_01JHun4BthW1irPjxB4N1Bhw


ctessum:
## Four checks are red on `main` already; none is this PR's

I checked the base branch before this PR's own CI landed, so the expected reds are on the record with their root causes. Run [33982406818](https://github.com/EarthSciML/EarthSciAST/actions/runs/33982406818) on `main` @ `3a01c56` — the commit this branch is cut from — fails these four, and each reproduces on a checkout that does not contain this diff:

| Check | Fails at | Root cause | In this diff? |
|---|---|---|---|
| `Lint and format (all languages)` | `ruff check` | `F401` — `from pathlib import Path` unused in `pkg/earthsci-ast-py/tests/test_conformance_round_trip.py:25` | no |
| `rust-tests (beta)` | `Run Rust tests` | `tests/performance.rs` does not compile | no |
| `rust-tests (1.89.0)` | `Run Rust tests` | same | no |
| `rust-tests (stable)` | `Build WASM` | fails before the test step is reached | no |
| `conformance-results` | roll-up | consequence of the above | no |

Two things worth spelling out, because both hid the failures from a local run:

**The lint job fails closed on step 1.** `ruff check` is the first of nine steps, so `cargo fmt`, `cargo clippy`, `eslint`, `gofmt`, `golangci-lint` and the version-sync check have all been **skipped**, not passed, for as long as that import has been unused. I ran each of them locally against the CI-pinned toolchains (`ruff 0.16.4`, Rust `1.98.0`) and this branch is clean on every one — but CI will not say so until the ruff step is green.

**The Rust reds are a feature-surface gap, not a test regression.** CI runs the test suite over the whole feature surface; `--no-default-features`-style local runs miss it. `cargo test --release` (default features) gives **1421 passed, 0 failed** here, but `cargo check --tests --features performance` reproduces the CI error exactly:

```
error[E0063]: missing field `default_units` in initializer of `earthsci_ast::Species`
   --> tests/performance.rs:526:9   (also 535, 544)
error[E0063]: missing fields `default_units`, `distribution`, `shape` and 1 other field
             in initializer of `earthsci_ast::extension::types::Parameter`
   --> tests/performance.rs:577:9
error[E0063]: missing fields `analyses`, `tests` and `tolerance`
             in initializer of `earthsci_ast::ReactionSystem`
   --> tests/performance.rs:584:18
```

`tests/performance.rs` is byte-identical on `origin/main` and on this branch. These are struct literals that were never updated as fields were added to `Species`, `Parameter` and `ReactionSystem`. This is the same rot the workflow's own comment already flags for the `cargo clippy` step:

> `performance` has ~14 call sites still passing `ExpressionNode` where `Expr::Operator` now wants `Arc<ExpressionNode>`.

— which has gone unseen because clippy sits behind the rustfmt gate, which sits behind the ruff gate.

## No fix exists to port, so here are the patches

There is no open PR or revert carrying a fix for either, so there is nothing for me to port into this branch. Both are one-liner-to-small and outside this PR's scope, so I am proposing rather than pushing them; say the word and I will land either or both here, or open a separate PR.

**Lint** — delete the unused import:

```diff
--- a/pkg/earthsci-ast-py/tests/test_conformance_round_trip.py
+++ b/pkg/earthsci-ast-py/tests/test_conformance_round_trip.py
@@ -22,7 +22,6 @@
 import json
 from collections.abc import Iterator
-from pathlib import Path
 from typing import Any
```

**Rust** — `tests/performance.rs` needs the five struct literals brought up to the current field sets (`Species.default_units`; `Parameter.default_units`/`distribution`/`shape` + one more; `ReactionSystem.analyses`/`tests`/`tolerance`). `..Default::default()` on each literal is the smaller change if those types derive `Default`; otherwise it is five explicit field additions. The `stable` leg's WASM failure is a third, separate thing and I have not root-caused it — it fails before the test step, so it is not related to either of the above.

I have not spent a re-run on any of these: they are deterministic compile and lint errors, not flakes, and a re-run would reproduce them identically.

## What this PR's own CI still owes

This PR's run was still queued when I wrote this. I expect it to reproduce exactly the four reds above and nothing else; I will follow up here once it completes, and if anything red turns out to be mine I will fix and push rather than leave it.

---

ctessum:
## `Package Security Scan` failed — a fifth pre-existing red, and it has two independent causes

This one I did not cover in my previous comment, so here it is. `Package Security Scan` ([job](https://github.com/EarthSciML/EarthSciAST/actions/runs/33997753444/job/101391138614)) is **also red on `main` already** — the last five runs of `security-scan.yml` on `main` have failed, including run [33982406830](https://github.com/EarthSciML/EarthSciAST/actions/runs/33982406830) at `3a01c56`, this branch's base. It last went green at `de784f3` on 2026-09-04.

This PR changes **no dependency manifest at all** — `git diff origin/main --name-only` matches none of `package*.json`, `Cargo.*`, `Project.toml`, `pyproject.toml`, `go.mod`, `requirements*`. It adds a Julia source change, tests, fixtures, JSON goldens and Markdown, and nothing else.

The job log shows two separate failures stacked on each other:

**1. A real high-severity advisory in a transitive npm dependency.** Reproduced locally with `npm audit` in `pkg/earthsci-ast-ts`:

```
fast-uri  high  |  vulnerable range 3.0.0 - 3.1.5  |  fixAvailable: true
  GHSA-5jgf-p345-68v8  host confusion via skipped IDN canonicalization on scheme-relative references
  GHSA-f65p-4m7j-42xc  SSRF via malformed IPv6 normalization
  GHSA-fph4-wmhf-6fwf  SSRF via repeated hostname percent-decoding
  GHSA-jqff-g426-hqxp  host confusion via percent-encoded scheme normalization
```

The lockfile pins `fast-uri` 3.1.5, pulled in transitively by `ajv` (`node_modules/ajv -> ^3.0.1`). npm reports a fix is available, so this is a lockfile bump rather than a code change. It is a genuine finding and worth fixing — just not in a PR about Julia index resolution, so I am flagging rather than pushing it. **Say the word and I will do it here or in a separate PR**; given it is a live high-severity advisory, a separate PR that can land quickly seems better than parking it behind this one.

The scan also warns on `AGPL-3.0-only` not being in the allowed license list for `earthsci-ast-ts` and `earthsci-ast-rs`, and on an empty license string for `earthsci-ast-py`. Those are the repo's own packages, so that reads like the allowlist needs the project's own license added rather than anything being wrong with the packages — but I have not dug into the scanner config and am not proposing a change there.

**2. The job cannot post its own report, and that is what actually errors it out.** The final step 403s:

```
RequestError [HttpError]: Resource not accessible by integration
  POST https://api.github.com/repos/EarthSciML/EarthSciAST/issues/184/comments
  x-accepted-github-permissions: 'issues=write; pull_requests=write'
##[error]Unhandled error: HttpError: Resource not accessible by integration
```

The workflow's `GITHUB_TOKEN` lacks `issues: write` / `pull-requests: write`, so `actions/github-script` throws unhandled and fails the job. This means the scan report never reaches a PR — the findings above are only visible by opening the raw job log, which is presumably not the intent. Adding

```yaml
permissions:
  contents: read
  issues: write
  pull-requests: write
```

to the job in `.github/workflows/security-scan.yml` would fix the reporting; whether the job should also *fail* on a high-severity finding is a policy call for you, not something I should decide.

## No re-run spent

Neither cause is a flake — one is a vulnerability-database result and the other is a token-permission error — so a re-run would reproduce both identically. I have not spent one.

I will keep watching; the remaining checks on this PR are still queued or in progress and I will report on them once they land.

---

ctessum:
## Follow-up: the WASM failure, root-caused

In my first comment I listed `rust-tests (stable)` as failing at `Build WASM` and said I had not root-caused it. I have now, so here it is — it is the one red I left open, and it is not this PR's either.

**Cause 1 — the lib does not compile for `wasm32`.** `ProblemOptions.const_arrays` is declared

```rust
// pkg/earthsci-ast-rs/src/problem.rs:252
#[cfg(not(target_arch = "wasm32"))]
pub const_arrays: HashMap<String, ArrayD<f64>>,
```

but the f32 precision-rounding loop that reads it, ~900 lines below, carries no such gate:

```rust
// pkg/earthsci-ast-rs/src/problem.rs:1179
for (k, a) in opts.const_arrays.iter_mut() {
```

so on `wasm32` the field is absent:

```
    --> src/problem.rs:1179:28
     = note: available fields are: `p`, `u0`, `model_name`, `metaparameters`, `base_path` ... and 4 others
error[E0282]: type annotations needed
    --> src/problem.rs:1182:17
```

(the E0282 is fallout from the E0609, not independent). Reproduced locally with the CI invocation verbatim, `RUSTFLAGS='--cfg getrandom_backend="wasm_js"' cargo build --target wasm32-unknown-unknown --features wasm`. Gating the loop the same way its field is gated fixes it, and I verified that rather than assuming it — `cargo build --lib --target wasm32-unknown-unknown --features wasm` **succeeds** with:

```diff
--- a/pkg/earthsci-ast-rs/src/problem.rs
+++ b/pkg/earthsci-ast-rs/src/problem.rs
@@ -1176,6 +1176,7 @@
         for (k, v) in opts.u0.iter_mut() {
             *v = precision::of_variable(k).round(*v);
         }
+        #[cfg(not(target_arch = "wasm32"))]
         for (k, a) in opts.const_arrays.iter_mut() {
             let kp = precision::of_variable(k);
             if kp.is_f32() {
```

**Cause 2 — and this is why I am not calling that a fix for the step.** With the lib building, the same command then fails on the `esm` binary with 11 further errors: `src/bin/esm.rs` reaches for `PrepareProvider`, `BuildProviderFactory`, `run_pde_tests_filtered`, `ProblemOptions::build_pipeline` and `ProblemOptions::build_providers`, all of which are `cfg(not(target_arch = "wasm32"))` in the lib. The CLI is simply not a wasm target and carries no guard saying so, while the workflow step

```yaml
- name: Build WASM
  run: cargo build --target wasm32-unknown-unknown --features wasm --verbose
```

builds every target in the package, binaries included. Since the artifact the step exists to check is the wasm **library**, the smaller fix is probably `cargo build --lib --target wasm32-unknown-unknown --features wasm` — with the `problem.rs` gate above, that is green today. Gating the bin instead would also work. Which of those you prefer is a call for you; both are outside this PR.

So the WASM step needs two changes, not one, and I would rather say that than hand over a patch that turns one error into eleven.

## Status of the five reds on this PR

All predicted, all reproduced with the same root causes, none touching this diff:

| Check | Result | Same cause as `main`? |
|---|---|---|
| `Lint and format (all languages)` | ❌ | yes — the single `F401`, no findings in this PR's files |
| `rust-tests (beta)` | ❌ | yes — `performance.rs` E0063, 5 errors |
| `rust-tests (stable)` | ❌ | yes — `Build WASM`, now root-caused above |
| `Package Security Scan` | ❌ | yes — `fast-uri` advisory + the 403 |
| `rust-tests (1.89.0)` | pending | expected: same as beta |

`typescript-tests (20)` has passed. The Julia, Python, Go and remaining legs are still running; I will report the `julia-tests` result, which is the whole-suite number I still owe from the PR description, once those land. Nothing red so far is mine to fix, and if anything red turns out to be, I will fix and push rather than comment.

No re-runs spent: every one of these is a deterministic compile, lint, or advisory result.

---

ctessum:
## The Julia whole-suite result I owed — and CI is green everywhere it is green on `main`

The PR description said the full `Pkg.test()` sweep had not reported yet and that I would follow up. It has now, and so has almost all of CI.

**Local full Julia suite** — `julia --project=. -e 'using Pkg; Pkg.test()'` in `pkg/EarthSciAST.jl`, run to completion on this branch:

```
Test Summary:        |  Pass  Broken  Total      Time
EarthSciAST.jl Tests | 22584       7  22591  66m22.4s
     Testing EarthSciAST tests passed
```

**22584 passed, 0 failed.** The 7 `Broken` are the suite's pre-existing `@test_broken` markers, not failures. That includes this PR's two new Julia testsets, which I had already verified in isolation (`tree_walk_elementwise_obs_gather_test.jl` 17/17, `conformance_elementwise_observed_gather_test.jl` 62/62, both non-vacuous — they fail on a `git stash` of the source change).

**CI on this PR** — every check that is green on `main` is green here:

| | |
|---|---|
| ✅ `julia-tests` | 1.10, 1.11 pass (1.12 still running; I will note it if it does anything other than pass) |
| ✅ `python-tests` | 3.9, 3.10, 3.11, 3.12 |
| ✅ `go-tests` | 1.21, 1.22, 1.23 |
| ✅ `typescript-tests` | 20, 22 |
| ✅ `generate-docs` | |
| ❌ `Lint and format`, `rust-tests` (stable / beta / 1.89.0), `Package Security Scan` | the five reds that are red on `main` at this branch's base, each root-caused in the comments above |

So the complete picture across the three executing bindings, counting only what actually ran:

* **Julia** — 22584 passed / 0 failed, whole suite, including the new category.
* **Python** — `pytest tests/` 3135 passed / 0 failed locally; all four CI legs green.
* **Rust** — `cargo test --release` 1421 passed / 0 failed on default features locally, and the new conformance runner passes. The three CI legs are red for the pre-existing `performance.rs` and WASM breakage described above, neither of which this diff touches. **I want to be plain that this is a real gap in what I can claim**: because those legs never reach the test step, the whole Rust feature surface is unverified on this branch by CI, and my local number covers default features only. It is unverified on `main` for the same reason and by the same amount, so this PR does not make it worse — but "Rust is green" would be overstating it, and I am not going to say it.

Nothing red is this PR's, and nothing further is blocked on me. I will keep watching until this merges or closes.

---

PR #186, opened 2026-09-05
fix(julia): a §6.6.5 assertion may read an OBSERVED that no live equation consumes

### LINKED ISSUE #176: Julia: a DEAD observed is dropped from BuildInspection.observed_exprs, so a §6.6.5 assertion on it is unanswerable

ctessum:
Fixes #176.

## What this fixes

An inline test's natural target is a quantity computed **for the test** — a
tendency, a flux, a diagnostic — which by construction nothing else reads.
Julia refused to answer one:

```
array state 'diag' has no cells in var_map
```

on a document whose defining equation it had loaded, and which both the Rust
and the Python runners evaluate.

The cause is not the assertion path but what reaches it. The elementwise
array-observed fold (`_fold_elementwise_array_observeds`) inlines such an
observed into its readers and **drops its equation**; a DEAD one has no readers
to be inlined into, so it is dropped outright and reaches neither
`BuildInspection.observed_exprs` nor `observed_defs`. `_observed_field` found
no body and fell through to the state-cell error above. esm-spec §6.6.5 admits
any shaped variable and §5.23 makes a reference denote its expansion — neither
conditioned on the dynamics reading it — so on Julia an author had to wire a
diagnostic into the dynamics, *changing the model*, to make a test runnable.

## What is in the change

`_observed_field` now reads an observed from three sources, in order, so it is
readable however the build chose to treat it:

1. the **published body** (`observed_exprs` / `observed_defs`) — unchanged;
2. the **const-array buffer**, for an observed the build materialized. A
   document-literal array (`base = const([1,2,3,4])`) was unassertable for the
   same reason: the build folds it into the const-array registry and publishes
   no body for it;
3. the component's **own defining equation** (`_authored_observed_body`) — the
   #176 case.

The authored body is a whole-array expression (`2 * base`) while
`evaluate_cellwise` walks one output cell at a time, so it is lifted to the
per-cell `arrayop` form through the same `_index_array_leaves` the build's own
promotion uses — §4.3.4 name alignment holds, so a `[x]`-shaped operand in an
`[x,y]` result replicates along `y` instead of being gathered positionally —
and its operand names are rewritten into the build's flattened scope (`base` →
`M.base`). Producers the body names resolve from the published graph, or
recursively from their own authored bodies, one level of deadness down. The
recursion guard is the dependency **chain**, not the visited set: two siblings
reading the same dead observed are a diamond, not a cycle, and each must
resolve it.

Nothing that answered before changes: an observed the build publishes keeps its
published body, and a name the component does not declare is still refused with
the same message.

## Cross-binding status

Measured on `a1dc9bb`-era `main` with the new fixture (a dead `diag = 2*base`
and a dead `chain = diag + base`):

| Binding | Before | After |
|---|---|---|
| **Julia** | `array state 'diag' has no cells in var_map` | 5/5 assertions PASS |
| **Python** | 5/5 PASS (evaluates the ordered observed graph on demand) | unchanged |
| **Rust** | 5/5 PASS (requests the observed from the runtime) | unchanged |

## Tests

* **New conformance category** `tests/conformance/pde_inline_dead_observed/` —
  fixture, Julia-minted golden, and adapters for all three executing bindings
  (`conformance_pde_inline_dead_observed_test.jl`,
  `test_pde_inline_dead_observed_conformance.py`,
  `pde_inline_dead_observed_conformance.rs`), each gating every assertion
  against **both** the golden actual and the fixture's own declared `expected`.
  The fixture's only state is integrated with a zero right-hand side, so the
  trajectory is constant and the goldens are integrator-independent: a
  divergence there is a semantics divergence, never an integrator one.
* **Unit coverage** `pkg/EarthSciAST.jl/test/pde_inline_dead_observed_test.jl`
  — the rank-2 lift, §4.3.4 broadcast alignment through the lift (a positional
  gather would read `m[i,i]` and is caught by two `coords` samples), the
  diamond, the materialized-buffer path, that a published body still wins, and
  that an undeclared name is still refused. Every expectation is an independent
  hand-computation from the fixture's constants. Verified non-vacuous: the
  whole dead-observed testset fails on a clean checkout of the source file, and
  the diamond cases fail specifically if the recursion guard is the visited set.
* **ESM_COMPLIANCE_VALIDATION_MATRIX** row BEHAV-06-B-009.

## Test results

* **Julia** — new suites: `48 passed` (unit) and `40 passed` (conformance
  adapter). Regression sweep over the neighbouring suites, all green:
  `pde_inline_tests_test.jl` (94), `pde_inline_scalar_slot_collision_test.jl`
  (15), `observed_field_static_test.jl` (19),
  `observed_materialization_test.jl` (20), `build_inspection_test.jl` (24), and
  the three existing `conformance_pde_inline_*` adapters (154).
* **Python** — `pytest tests/test_pde_inline_dead_observed_conformance.py
  tests/test_pde_inline_tests.py`: `22 passed`. `ruff check` and
  `ruff format --check` clean on the new file.
* **Rust** — `cargo test --test pde_inline_dead_observed_conformance --test
  pde_inline_ic_param_override_conformance`: `3 passed, 0 failed`.
  `cargo fmt --check` reports nothing on the new file (the repository's
  pre-existing `fmt` findings elsewhere are unchanged).

Full `Pkg.test()` was not run here: the container has no Julia toolchain by
default and the test target's MTK/Catalyst stack was out of reach, so the
Julia coverage above is the affected suites run individually against a
`develop`ed checkout.

## Not in scope

The **state-dependent** half of #176's reproduction is separate and is not
fixed here. On `main` no binding answers a state-dependent array observed —
dead or live: Julia and Rust report `has no cells in var_map` and Python the
same, because a state-dependent observed is neither an ODE row nor a
build-time product. That is what #177 addresses. With this change a
state-dependent dead observed now reaches the evaluator and reports
`E_TREEWALK_UNBOUND_VARIABLE: u` — exactly the pre-#177 behaviour for a *live*
one — so the two fixes compose: #177 puts the trajectory sample into the scope
`_observed_field` builds, and this one makes a dead observed reach that scope
at all.

https://claude.ai/code/session_0192jFJzkmp4hN6Fc6j64eDV

---
_Generated by [Claude Code](https://claude.ai/code/session_0192jFJzkmp4hN6Fc6j64eDV)_


ctessum:
## Merged with `main`, four review findings fixed

Pushed to this branch (fast-forward from `4f7b33a`), so the PR description above still describes the change; this is what moved since.

### Merged with `main` (through `8b8f276`) — three conflicts, all real

* **`pde_inline_tests.jl`** — #177 landed `_state_scope` and threaded the trajectory sample through `_observed_field` in the same lines this PR refactored. Both are kept, and they are now **wired together**: `_authored_observed_body` receives the state scopes, so its bare-name qualification covers state stems (`u` → `M.u`). Without that, the composition this PR's "Not in scope" section predicts would silently not happen — a dead *state-dependent* observed would still die on `E_TREEWALK_UNBOUND_VARIABLE: u`. It now answers, and a new testset pins it. Non-vacuous: dropping the state scopes from the qualification step fails 8 of that testset's 9 expectations while every other testset in the file still passes.
* **`ESM_COMPLIANCE_VALIDATION_MATRIX.md`** — `main` now claims BEHAV-06-B-009 and ‑010, so this row is **BEHAV-06-B-011**.
* **`runtests.jl`** — #205's conformance include lands on the same line; both kept.

### Review findings, all fixed

1. **Reuse.** The per-cell lift duplicated `shape_promotion.jl`'s `_lift_to_arrayop` — same `_p<k>` loop convention, same `_AxisAlign` wiring, same arrayop wrapper — differing only in that ranges are resolved `[1, extent]` literals rather than index-set references. That difference is now a `bounds` keyword on the build's own lift, and the assertion path calls it. Two copies of the lift cannot drift apart when there is one.
2. **Qualified-first const arrays.** The producer skip and the materialized-buffer read both fell back to a bare key, so in a build where a *sibling* component's array is bare-keyed they would read that sibling's numbers — the hazard BEHAV-06-B-010 hardened for state cells. `_const_array_key` now applies the same two-pass rule: the model-qualified key wins, and a bare key is this component's only when no model qualifies that suffix. Pinned by a two-component fixture (`M1.base = [1,2,3,4]`, `M2.base = [100,…,400]`) asserted from both sides.
3. **The changed diagnostic, documented and pinned.** A declared observed whose body cannot be evaluated at assertion time now reports `E_TREEWALK_UNBOUND_VARIABLE: <name>` where it previously reported `array state '<v>' has no cells in var_map`. Same ERROR verdict, better message — but a visible change, so it is now stated in the source, in CONFORMANCE_SPEC §5.27.3 and in the matrix row, and pinned by a test.
4. **CONFORMANCE_SPEC §5.27.3.** The category had a manifest, a golden and three adapters but no normative section, unlike its neighbours §5.27.1/§5.27.2. Written as the rule (an observed is readable however the build chose to treat it — three sources in order; the recursion guard is the dependency chain, not the visited set) plus the gate. §5.27.2's closing paragraph no longer calls the deadness divergence untracked.

### The Julia coverage the description could not run is now run

`pde_inline_dead_observed_test.jl` **76 passed**; `conformance_pde_inline_dead_observed_test.jl` **40**; `conformance_pde_inline_observed_state_dependent_test.jl` **52** (#177's, the suite most exposed to this resolution). Regression sweep, all green: `shape_promotion_test.jl`, `shape_promotion_consumer_refs_test.jl`, `pde_inline_tests_test.jl`, `pde_inline_scalar_slot_collision_test.jl`, `observed_field_static_test.jl`, `observed_materialization_test.jl`, `build_inspection_test.jl`, and the `pde_inline_observed_rank2` / `_param_rank2` / `_ic_param_override` conformance adapters. Python **27 passed**, Rust **2 passed**; `ruff check`, `ruff format --check`, `cargo fmt --all --check`, `cargo clippy --all-targets -D warnings` and `check-version-sync.sh` all clean.

### Why CI was red, for the record

None of it was this PR's code — all four failures were the old base's, and `main` has since fixed each:

| Job | Cause |
|---|---|
| `rust-tests` ×3 | The Rust lib did not compile: `error[E0609]: no field 'const_arrays' on type 'ProblemOptions'` |
| `standard-conformance-testing` → `conformance-results` | Skipped because rust-tests failed, so the results job posted a failure comment it lacked permission to post |
| `Lint and format` | `F401 pathlib.Path imported but unused` in `tests/test_conformance_round_trip.py`, a file this PR does not touch |
| `Package Security Scan` | Failed in its own comment step |

The merge above is the fix; CI is re-running on the new head.

---

PR #202, opened 2026-09-06
fix(bindings): resolve override keys forward; pin `wrt` and dimension/parameter clashes in §6.6.5

ctessum:
## Summary

Review fixes on top of #179, which this branch carries — #179's commit is included here, so landing this lands both. Four findings from reviewing that PR, plus the merge with `main`.

### 1. Python resolved override keys backwards, applying one override to two parameters

`_resolve_override` matched *every key of which the parameter name is a dotted suffix*:

```python
longer = sorted(k for k in overrides if k.endswith("." + name))
```

With the flattened parameters `Left.solo` and `Right.Left.solo` both in a build, the override `{"Right.Left.solo": 9.0}` is an exact hit on the second **and** a backwards suffix match on the first, so both got 9.0. Julia's `_canonicalize_override_keys` and Rust's `canonicalize_override_keys` map each key *forward* to the one name it designates and were unaffected — so this was also a cross-binding divergence in the code that §6.6.2 rule 2 governs.

`_resolve_override` now takes the build's name set and does the forward resolution: a key that is itself a known name is never read as a more-qualified spelling of another name. All three call sites (scalar parameters, scalar initial conditions, array parameters) pass their name set; `known` defaults to `{name}`, so a caller resolving one isolated name behaves as before, and `name` is always folded into the view so a partial set cannot make a parameter unresolvable. Affects `initial_conditions` as well as `parameter_overrides`.

### 2. Duplicate CONFORMANCE_SPEC section number

#179 numbered its new section `### 5.25`, which already existed (*Inline-Test Build Reuse and Test Selection*, with subsections `5.25.1`–`5.25.4` cited from BEHAV-12), and its nested gate heading `#### 5.25.1`. With `main`'s `§5.26` (`enums` member domain) and `§5.27` (array-observed assertions) now landed, the two sections from this line are **§5.28** and **§5.29**, and every citation of them follows.

### 3. Julia treated a `wrt` as a free mention

Julia decided the free mention with `free_variables`, which adds a node's `wrt` to the free set *after* binder subtraction; Rust's `mentions_free` and Python's `_mentions_free` ignore `wrt` entirely. A reference containing `deriv(u, wrt: "x")` was therefore wrapped in Julia alone — the divergence the new section exists to close. Julia gets its own `_mentions_free` mirroring the other two, and §5.28 now states normatively that a `wrt` is a differentiation target, not a free mention.

### 4. A dimension name could silently shadow a parameter

The wrap was decided on the name alone. If a parameter's bare alias equalled a `shape` index-set name, a reference that previously read the parameter now read the cell's 1-based index instead — the same expression, a different number, no diagnostic. That contradicts the "nothing that evaluated before evaluates differently" invariant all three docstrings assert.

`bind_dimension_names` now takes the reference's build-time parameter scope and rejects the clash with an error naming it, in all three bindings. A reference that does not mention the name, and a gather that rebinds it as its own loop symbol, are unaffected. A fault rather than a precedence rule because either choice would be a guess about a document where one name means two things in one scope; §5.28 says so and the matrix row records it.

## Spec and conformance

- `CONFORMANCE_SPEC.md` §5.28 gains the `wrt` clause and the scope-clash clause; §5.28.1 names the per-binding gates for both, since neither can be expressed as a passing fixture. §5.29 is the override-key section, renumbered.
- `ESM_COMPLIANCE_VALIDATION_MATRIX.md` — this line's rows are **BEHAV-06-B-011** and **-012** (`main` landed -009 / -010 for the array-observed work), with -011 extended to state both new clauses.
- New unit coverage in all three bindings: forward key resolution including the two-parameter case that motivated it, deterministic choice when two keys designate one name, the `wrt` exclusion, and the scope clash.

## Merge with `main`

Two conflicts, both from `main` claiming identifiers this branch had also taken — the matrix rows (resolved as above) and `test_pde_inline_tests.py`, where both sides appended independent test functions at the same point (kept both, no name collisions).

## Verification

Full suites on the merged tree:

| | result |
|---|---|
| Rust `cargo test --release` | 124 targets, 1427 passed, 0 failed |
| Python `pytest tests/` | 3148 passed, 25 skipped, 3 xfailed |
| Julia `Pkg.test()` | 22626 passed, 0 failed, 7 broken (pre-existing) |

`ruff check src/` reports the same 5 findings as the baseline, all in files this branch does not touch.

https://claude.ai/code/session_01TgVneGRbaX4guxxUpJx5Kp

---
_Generated by [Claude Code](https://claude.ai/code/session_01TgVneGRbaX4guxxUpJx5Kp)_


ctessum:
## `conformance-results` failure on `7cd860d` — not this PR's, and not a conformance failure

`conformance-results` ([job](https://github.com/EarthSciML/EarthSciAST/actions/runs/34039834374/job/101506864910)) went red on the superseded head `7cd860d`. No conformance test ran and failed. Two independent causes, neither in this diff:

**1. Its upstream job was cancelled, not failed.** The workflow's `concurrency` group (`conformance-${{ github.ref }}`) cancels superseded runs. Pushing `b4e4e3a` at 14:56:22 superseded run 815, and every one of its still-queued jobs — `standard-conformance-testing` included — was cancelled at exactly that timestamp. `conformance-results` is `if: always()`, so it ran anyway, found no results artifact, and reported "no results generated". That message means "the run was cancelled", not "conformance broke".

For the record, the one test job that had finished before the cancellation, `python-tests (3.10)`, **passed**.

**2. Its comment step 403s on every PR.** The step then failed outright:

```
POST /repos/EarthSciML/EarthSciAST/issues/202/comments
403 Resource not accessible by integration
x-accepted-github-permissions: issues=write; pull_requests=write
```

`.github/workflows/conformance-testing.yml` declares only:

```yaml
permissions:
  contents: read
```

so `GITHUB_TOKEN` cannot post the results comment. This is repo-wide and pre-existing — it will fail on any PR that reaches this job, whatever the conformance outcome, and it is independent of this branch. #184 reports the same 403 on the same job.

**Proposed patch** (deliberately *not* applied here — it is a workflow-permissions change unrelated to this PR's subject, and folding it in would widen the diff):

```yaml
permissions:
  contents: read
  pull-requests: write
```

`issues: write` would also satisfy the API, but `pull-requests: write` is the narrower grant that covers this step. I have not found an open PR fixing it, so there is nothing to port; happy to open a separate one-line PR if that is wanted.

No re-run of run 815 — it is superseded and its jobs are cancelled, so a re-run would tell us nothing. The current head `b4e4e3a` has a fresh run ([816](https://github.com/EarthSciML/EarthSciAST/actions/runs/34040724464)) with all 16 jobs queued; the lint failure that was red on the previous head is fixed there. I will keep watching and act on anything that comes back red on that run.

---

PR #212, opened 2026-09-07
fix(validate): reject a declaration named after the independent time variable

### LINKED ISSUE #200: validate accepts an observed named `t`, which then collides with the time variable at build time
## What happened

A model declared an observed variable named `t` (a fuel time-lag constant, in a WRF-Fire fuel-moisture component). `esm validate` accepted the file. At build/test time the name collided with the independent time variable:

- bare model: `Observed variable P.t has no defining expression`
- with a subsystem mounted: other equations that read `t` silently received the simulation time instead of the observed (`log(t) = -inf` at t = 0).

Renaming the observed fixed it. Repro: `probe_observed_named_t.esm` (available on request; a one-observed model suffices).

## Expected

`esm validate` should reject a variable, parameter, or observed named `t` (and any other reserved independent-variable name the spec defines), or the spec should state that `t` is reserved and the schema should enforce it. Silently shadowing the time variable is the worst of the three behaviours.

Rust CLI built from the `inline-test-array-observeds` branch (PR #177).

https://claude.ai/code/session_015LxnQcZYdcu9qGuUkncwvc

ctessum:
Fixes #200

## The rule

A **declaration** spelled with a globally-scoped name is a hard structural error, reported at the offending key.

**Reserved set — `{ domain.independent_variable (default "t"), "_var" }`.** It **follows the document**, it is not hard-coded to the literal `"t"`: a file that renames its independent variable to `s` reserves `s` and **frees** `t`, which is then an ordinary name. Both halves are pinned in the shared corpus (`tests/invalid/reserved_variable_name_renamed_independent.esm` and its valid twin `tests/valid/independent_variable_renamed.esm`), so a binding that hard-codes `"t"` fails one of the two.

**Declaration sites — all three maps that populate a component's symbol namespace:**

| Site | Pointer |
|---|---|
| `models[M].variables` | `/models/<M>/variables/<name>` |
| `reaction_systems[S].species` | `/reaction_systems/<S>/species/<name>` |
| `reaction_systems[S].parameters` | `/reaction_systems/<S>/parameters/<name>` |

A species and a reaction parameter become symbols of the derived ODE system exactly as a `variables` entry does (§7.4), so all three collide identically.

**Not reserved: spatial coordinate names.** `x`, `y`, `z`, `lon`, `lat`, `lev` resolve as coordinates only in a *coordinate position* (§11.4) — `tests/valid/units_dimensional_analysis.esm` declares `x` as an ordinary position variable, and it must keep validating.

## Error code and severity

`reserved_variable_name` — **hard error** (`is_valid: false`), the same code name in all five bindings, with `details: {"name": …, "reserved_as": "independent_variable" | "operator_placeholder"}`.

Hard, not a warning, because the collision is **provable, deterministic and statically decidable from the declaration map alone** — this repo's established policy for exactly that class (the units severity contract, CONFORMANCE_SPEC §7.1.1). And because of what it cost:

1. the document **validated**;
2. built bare, the model reported the observed as having *no defining expression* — a diagnostic about the wrong thing;
3. built with a subsystem mounted, every equation that read `t` **silently received the simulation time**, so `log(t)` was `-inf` at `t = 0` and every number downstream was finite, plausible and wrong.

The declaration never wins: the independent variable and `_var` are implicitly declared in every component's expression scope (§4.9.1) and are resolved **by name ahead of the declaration maps** in every binding, so the implicit symbol shadows the declaration rather than the other way round. Rejecting costs an author nothing — a declared name is a free choice, so the fix is to spell it anything else.

Reproduced before the fix (Python, a one-observed model named `t`): `validate` returned `is_valid=True`; then `evaluate_rhs` divided by zero, warned *"unrecognized algebraic equation with LHS 't' was not applied to the ODE RHS; any state it constrains stays frozen at its initial value"*, and raised `Non-finite derivatives`. After the fix: `reserved_variable_name @ /models/P/variables/t`.

## Schema: deliberately unchanged

Considered and rejected. The reserved name is the **value of another field** (`domain.independent_variable`), which JSON Schema cannot express; a hard-coded `propertyNames: {not: {const: "t"}}` would reject a conforming document that renames its independent variable and would say nothing about the renamed one. A schema rejection would also **pre-empt the layer where the finding is pinned** — a schema-invalid file never reaches structural validation (CONFORMANCE_SPEC §7.1.2), so the `(code, path)` pin would be unsatisfiable. The rule is a structural check in all five bindings.

## Per-binding summary

| Binding | Where | Note |
|---|---|---|
| **Julia** | `src/validate.jl` — `_reserved_declaration_names` / `_check_reserved_declaration_names!` / `_check_reserved_model_names!`, wired as step 3e of `validate_structural`; `RESERVED_VARIABLE_NAME` in `src/error_codes.jl` | recurses into subsystems |
| **Python** | `src/earthsci_ast/structural_checks.py` — `_reserved_declaration_names` / `_check_reserved_declaration_names`, `collect("reserved_variable_name", …)`; `ErrorCode.RESERVED_VARIABLE_NAME` in `error_handling.py` | this layer runs inside `load()`, so `load_string` rejects too |
| **TypeScript** | `src/validate/model-checks.ts` — `reservedDeclarationNames` / `validateReservedDeclarationNames`, wired into `validate/orchestrator.ts` for models, subsystems and reaction systems; `ERROR_CODES.RESERVED_VARIABLE_NAME` | recurses into subsystems |
| **Rust** | `src/structural.rs` — `reserved_declaration_reason` / `check_reserved_declaration_names`, called from `validate_model` and `validate_reaction_system`; `StructuralErrorCode::ReservedVariableName` + `codes::RESERVED_VARIABLE_NAME` + the pinned vocabulary vec in `diagnostic.rs` | typed layer, so the `(code, path)` conformance record is produced |
| **Go** | `pkg/esm/validate.go` — `reservedDeclarationNames` / `validateReservedDeclarationNames`, called from `validateModel` and `validateReactionSystem`; `ErrorReservedVariableName` in `codes.go` | |

Docs: `esm-spec.md` §4.9.1.1 (new — and the **normative home** of the reserved set, which the sibling `reserved_index_symbol` binder rule now reads rather than restating) and the §11.3 `independent_variable` row; `CONFORMANCE_SPEC.md` §7.1 code table; `ESM_COMPLIANCE_VALIDATION_MATRIX.md` BEHAV-04-G.

## Corpus changes — yes, seven documents had to change

**Three shared fixtures carried a redundant `t` parameter** declared purely as a "time placeholder", which the new rule (correctly) rejects. All three now rely on the §4.9.1 implicit declaration, and **all three still emit exactly the findings they pinned before** (verified directly in Python and Julia; Go's corpus-pin test and the TS conformance sweep are green):

- `tests/valid/units_dimensional_analysis.esm` (stays valid)
- `tests/invalid/units_invalid_derivative.esm` (still `unit_inconsistency` @ `/models/BadUnitsModel/equations/0`)
- `tests/invalid/units_gradient_operator_mismatch.esm` (still `unit_inconsistency` @ `/models/SpatialModel/variables/bad_sum`)

**Three binding-local ERA5 stubs were this bug in the wild.** The ERA5 short name for air temperature *is* `t`, so `pkg/{earthsci-ast-rs,earthsci-ast-go,earthsci-ast-py}/…/data_sources/era5.esm` each bound a temperature forcing channel to a parameter literally named `t` — a channel that could never be read, because every reference to `t` resolves to the clock. They now declare `air_temperature` with `from.file_variable: "t"` (the on-disk name is unchanged). Two call sites moved with them: `earthscidata_loader_coverage.rs` and `test_data_sources_cache.py`.

**One Python unit test** (`test_simulation_model_serialization`) built its exponential-decay model with the same redundant `t` placeholder; dropped (the model's balance is unchanged — one unknown, one equation).

A scan of all 879 JSON documents in the repo, plus a scan of all five bindings' test sources for inline documents, finds **no other** declaration of a reserved name — only the four new invalid fixtures.

## New fixtures

- `tests/invalid/reserved_variable_name_observed.esm` — the reported shape
- `tests/invalid/reserved_variable_name_parameter.esm` — both reserved names, incl. the ERA5 `t`
- `tests/invalid/reserved_variable_name_species.esm` — both reaction maps
- `tests/invalid/reserved_variable_name_renamed_independent.esm` — the rename reserves `s`
- `tests/valid/independent_variable_renamed.esm` — the rename frees `t`

plus entries in `tests/invalid/expected_errors.json`, which `compare-conformance-outputs.py` requires for every `tests/invalid/**` fixture ("fixture pins no (code, path); its rejection is unverifiable").

## What I tested

**I did NOT run `./scripts/test-conformance.sh` — CI does that.** Everything else below is a local run.

- **Python** — the full suite: **3177 passed, 19 skipped, 3 xfailed**. (I started with targeted runs; CI then surfaced the one remaining `t`-placeholder unit test, so I ran the whole suite to be sure there were no others.) `ruff check --config ruff.toml` clean. 5 new cases in `TestReservedDeclarationNames`.
- **TypeScript** — `tsc --noEmit` clean; the full suite: **2554 passed, 70 files**. 5 new cases in `validate.test.ts`; the `conformance.test.ts` sweep is what exercises the new fixtures and the three edited corpus files.
- **Go** — `go build ./...`, `go vet ./...`, `gofmt -l` all clean; the full `go test ./...` green. It caught that the new exported `ErrorReservedVariableName` needed an `api-surface.json` entry. 3 new tests in `validate_test.go`.
- **Rust** — `cargo check --lib` and `cargo fmt --check` clean; `cargo test --test structural_validation` (**41 passed, 0 failed**, incl. the 5 new cases) and `cargo test --test earthscidata_loader_coverage` (4 passed — the renamed ERA5 channel).
- **Julia** — `test/structural_validation_test.jl` (**136 pass**, incl. the new `reserved_variable_name` testset), `test/error_hierarchy_test.jl` (24), `test/units_fixture_consumption_test.jl` (122), `test/units_test.jl` (285). Plus a direct `validate_path` over all five new fixtures and the three edited ones, giving exactly the pinned findings.

The full Julia and Rust suites are left to CI.

## One note for the reviewer

I did **not** add the load-time (raw-JSON) rejection in Rust that `reserved_index_symbol` uses, so in Rust `load_string` still accepts such a document and only `validate()` rejects it. That keeps the conformance record shape identical across the five bindings (`schema_errors: []`, structural findings with paths), which is what §7.1.2 pins. Python's structural layer happens to run inside `load()`, so Python rejects at load as well; the emitted `(code, path)` record is the same either way. If you would rather Rust fail closed at load too, that is a small follow-up in `parse.rs` alongside the existing `reject_reserved_index_symbols`.


ctessum:
## CI status — the red is pre-existing on `main`, not from this PR

Conformance run [34080660118](https://github.com/EarthSciML/EarthSciAST/actions/runs/34080660118) on this branch:

| Job | Result |
|---|---|
| go-tests (1.21 / 1.22 / 1.23) | ✅ |
| julia-tests (1.10 / 1.11 / 1.12) | ✅ |
| python-tests (3.9 / 3.10 / 3.11 / 3.12) | ✅ |
| typescript-tests (20 / 22) | ✅ |
| Lint and format (all languages) | ✅ |
| rust-tests (stable / beta / 1.89.0) | ❌ |
| conformance-results | ❌ (downstream of rust-tests) |

**The Rust failure is not a test failure and is not from this change — it is a compile error that already exists on `main`:**

```
error[E0062]: field `analyses` specified more than once
   --> tests/performance.rs:605:9
605 |         analyses: None,
    |         ^^^^^^^^ used more than once
error: could not compile `earthsci-ast` (test "performance") due to 3 previous errors
```

`pkg/earthsci-ast-rs/tests/performance.rs` builds a `ReactionSystem` literal with `analyses: None` written twice (lines ~596 and ~605). The file is gated behind the `performance` feature, so it only compiles under CI's `--all-features` — which is why a default-feature `cargo test` passes locally.

Two things confirm it is pre-existing:

- `git diff origin/main -- pkg/earthsci-ast-rs/tests/performance.rs` is **empty** — this PR does not touch the file, and `git show origin/main:…` has the identical duplicated field.
- Run [34046256280](https://github.com/EarthSciML/EarthSciAST/actions/runs/34046256280) on **`main`** (`79c7ea617`) fails with **exactly the same job set**: `rust-tests` × 3 and `conformance-results` failed, `standard-conformance-testing` and `performance-analysis` skipped.

I left it alone deliberately — it is unrelated to the reserved-name rule, several agents are working in this repo concurrently, and a drive-by edit to an unrelated Rust test file would both muddy this diff and risk a conflict. The fix is deleting one duplicate `analyses: None,` line, and it belongs in its own change.

Rust's own coverage for **this** PR was verified locally instead: `cargo check --lib` and `cargo fmt --check` clean, `cargo test --test structural_validation` 41 passed / 0 failed (including the 5 new `reserved_variable_name` cases), and `cargo test --test earthscidata_loader_coverage` 4 passed (the renamed ERA5 channel).

The green Julia jobs above are the coverage I could not complete locally, so between them and the local runs every binding is now verified.

PR #233, opened 2026-09-07
fix(julia)!: a species in no reaction gets no equation, not D(X,t) = 0

ctessum:
Julia emitted `D(X, t) = 0` for a reaction species touched by no reaction — the empty stoichiometric sum. The other four bindings omit the equation. Julia was a 4–1 outlier, and this makes it 5–0.

Found while reviewing #216, whose new `operator_compose` diagnostic surfaced the divergence as a partial-merge warning on `tests/valid/full_coupled.esm`. That warning is a true report about this, not about the fixture; this PR is what clears it.

## The split

| Binding | species in no reaction | site |
|---|---|---|
| **Julia** | emitted `D(X) = 0` | `flatten.jl` — `rhs = if isempty(terms) NumExpr(0.0)` |
| **Python** | omits | `reactions.py` — `if species_rates[name] != 0` |
| **Rust** | omits, **deliberately** | `reactions.rs` — `if rate_terms.is_empty() { continue; }` |
| **Go** | omits | `flatten.go` — `if isNumericZero(rates[name]) { continue }` |
| **TypeScript** | omits | `flatten.ts` — `if (numericValue(rates[name]) === 0) continue` |

Rust does not merely happen to omit; it argues the case and cites the operative rule:

> A species with NO net contribution from any reaction gets NO equation, rather than a vacuous `D(X, t) = 0`. Emitting the zero equation makes such a species an ODE state that never moves… omitting it leaves the species an unknown that no equation names, which esm-spec §6.3.1's partition-by-complement classifies ALGEBRAIC.

## Why omit is right

The empty sum really is zero, so emitting it is defensible *arithmetic*. What settles it is that the two readings give the species different **roles**, and §6.3.1 classifies by complement:

- **With** the zero equation, `X` is a DIFFERENTIAL unknown — a state the solver advances whose derivative happens to be zero. It joins the `u` vector and is integrated.
- **Without** it, `X` is an unknown no equation names — §6.3.1's ALGEBRAIC category.

The second is what the species *is*. A declared species no reaction touches is not a quantity whose dynamics are known to be trivial; it is one the reaction network says nothing about, and §6.3.1 has a category for exactly that. Emitting the zero also makes `equation_count` depend on how many inert species a document happens to declare.

## The change

`lower_reactions_to_equations` (`pkg/EarthSciAST.jl/src/flatten.jl`) skips a species whose term list is empty, alongside the existing reservoir skip. Two lines plus the reasoning.

## Spec: the rule lands in §4.6.1

Two sections could have hosted it, and I picked the one §4.7.5 step 1 **defers to** — step 1 says "as specified in Section 4.6 `derive_odes`", so a rule about what `derive_odes` emits belongs where `derive_odes` is specified. Putting it only in §4.7.5's field-set table would make that table the normative home for a lowering rule, which is the wrong shape: the table describes a *consequence*.

- **New §4.6.1 "Which species `derive_odes` emits an equation for"** — states both exceptions (reservoir, and no-reaction), gives the §6.3.1 argument above, and notes that a later coupling rule may still define the species, so `derive_odes` is not deciding its final classification.
- **§4.7.5's `state_variables` row corrected.** It said the vector includes "every reaction-system species, **which gets a derived `D` equation at step 1**" — flatly true only if you emit the zero, and it is what misled this binding. It now names the exceptions, points at §4.6.1, and adds that such a species is still *in* the map: what changes is which half of the solved-for vector it lands in.

## `_FC_DIVERGENCES` disposition: **narrowed, not removed** — and here is why

The entry `"full_coupled" => ["equations"]` **stays**. Its cause shrinks from two to one:

| Equation | Before | After |
|---|---|---|
| `D(OH)` | **differed** — Julia `0 + <transport>` at position 4; corpus `<transport>` at position 1 | ✅ **byte-identical to the corpus**, position 1 |
| `D(NO2)` | matched | ✅ matches |
| `D(O3)` | differed | ⚠️ still differs — corpus `-1 * 1.8e-12 * …`, Julia `-1.8e-12 * …` |
| `D(NO)` | differed | ⚠️ still differs — same rendering |

Also now matching exactly: `equation_count` (4), `algebraic_variables` (`[]`), `state_variables`, and the **equation order**, which the `OH` divergence was shifting.

What survives is only the `-1 * rate` vs `-rate` coefficient rendering — a `to_ascii`-level difference in how the two lowerings build a −1 stoichiometry term, which the ledger already documents as its own general bullet ("`equations` on the reaction-lowering cases"). So `full_coupled` is no longer a case of its own; it is one of those.

I could not narrow it *structurally*: the ledger's finer syntax addresses one field of a variable-list key (`"parameters.update_kinds"`), and there is no way to say "equations 2 and 3 only". So the narrowing is (a) the documented cause reduced to one, with the `OH` half moved to a RESOLVED block, and (b) **a local pin that now asserts exactly what still differs** —

```julia
@test eqs[1] == corpus_eqs[1]                       # OH  — exact
@test eqs[4] == corpus_eqs[4]                       # NO2 — exact
for i in (2, 3)                                     # O3, NO — only `-1 *`
    @test eqs[i] != corpus_eqs[i]
    @test replace(corpus_eqs[i], "-1 * " => "-") == eqs[i]
end
```

— so the entry cannot silently widen back out. If the rendering is reconciled later, the entry goes entirely and this pin is what tells you it is safe to delete.

## Goldens that shifted

**No shared-corpus golden moved.** `tests/conformance/flatten/cases.json` is generated from the Python oracle, which is unchanged; Julia moved *toward* it. Two Julia-local pins shifted, both of which were pinning the artifact rather than the behaviour:

**1. `flatten_conformance_test.jl` — "inert reaction species"** (renamed *zero tendency, then transported* → *no tendency, transported only*).

| | Before | After |
|---|---|---|
| equation order | `[O3, NO, NO2, OH]` | `[OH, O3, NO, NO2]` |
| `OH`'s RHS | `0 + (-Transport.u) * grad(OH) + …` | `(-Transport.u) * grad(OH) + …` |

**Why the new value is correct:** `OH` gets no reaction equation, so §4.7.1 step 3's `_var` placeholder clone has nothing to merge onto and step 5 preserves it unmatched. The `OH` equation is therefore Transport's own clone, sitting in `models` order ahead of the reaction system — which is precisely what the shared corpus records and what the other four bindings produce. The leading `0 +` is gone because there is no longer a zero to sum onto. `algebraic_variables` stays `[]`: the surviving transport clone *does* name `OH`, so it is differential here — the algebraic case is the one with no operator at all, pinned separately.

**2. `flatten_test.jl` — "8a. couple additive connector transform (esm-spec §10.3)"**

| | Before | After |
|---|---|---|
| `D(Chem.A)` RHS | `0 + (-Sink.k) * Chem.A` | `(-Sink.k) * Chem.A` |
| assertion | `@test _has_op(eq_A.rhs, "+")` | `@test !_has_op(...)` + exact string |

**Why the new value is correct:** the testset builds a species in no reaction, so with §4.6.1 in force it has no tendency — which is exactly the case §4.7.2 spells out for `additive`:

> If `to` has no tendency yet, `expression` BECOMES it (`D(to) ~ expression`) — an additive term against an absent tendency is well defined, because zero is the additive identity.

The old `+` was pinning the vacuous zero equation, not the coupling behaviour the testset exists to test. Every other assertion in it is unchanged and still passes. This one is worth a second look precisely because it is a *coupling* test that was quietly depending on the lowering.

## Coverage added

`reactions_test.jl` — "A species in NO reaction gets no equation, and is ALGEBRAIC": pins that the lowering emits no `D(Xe,t)`, that `Xe` is still an UNKNOWN (it is declared and solved for; only which half changes), and that through `flatten` it lands in `algebraic_variables` with no equation naming it. Verified directly:

```
states    : ["Chem.A", "Chem.B", "Chem.Xe"]
algebraic : ["Chem.Xe"]
equations : ["D(Chem.A)/Dt = -Chem.k * Chem.A", "D(Chem.B)/Dt = Chem.k * Chem.A"]
```

## Testing

Targeted only — **no `./scripts/test-conformance.sh` and no full suite; CI does that.** I ran every Julia suite that references the affected fixtures, `derive_odes`, `lower_reactions_to_equations`, or `algebraic_variables`:

| Suite | Result |
|---|---|
| `reactions_test.jl` | **140 passed** |
| `flatten_conformance_test.jl` | **2495 passed** (was 2489; the new pins) |
| `flatten_test.jl` | **174 passed** |
| `reaction_species_order_test.jl` | **29 passed** |
| `classification_test.jl` | **184 passed** |

The other four bindings are untouched — they already do this — so there is nothing to run there.

## Relationship to #216

Independent branches; no overlapping lines. #216 adds the `operator_compose` merge diagnostic that *found* this. Once both land, Julia and Python agree on `full_coupled.esm` including on the diagnostic: both emit the same partial-merge warning, for the same remaining reason (the `_var` clone for `OH` genuinely lands nowhere), rather than disagreeing about whether there is anything to report.

### DIFF
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -58,6 +58,7 @@ green and your program does something different. Audit these first.
 | Go | `Substitute` | Single-pass, no longer transitive. `Substitute("a", {a: b, b: c})` returns `"b"`, not `"c"`. | Chained renames that were silently mis-applied now apply correctly. Goldens change. |
 | Julia, Python, Rust | `solve` default tolerances | `reltol = 1e-4`, `abstol = 1e-6` everywhere — **looser** than Rust's and Python's old defaults. | Trajectory assertions fail. |
 | Julia, Python, Rust | a failed **build** | Raises instead of returning a failed result. | Code inspecting `result.success` for build failures never sees it; the exception propagates. |
+| Julia | `derive_odes` / `flatten` on a reaction system | A species whose net stoichiometry is zero in every reaction gets **no equation**, where Julia used to emit `D(X, t) = 0`. The other four bindings already omitted it. | `equations`, `equation_count` and equation ORDER change; a `couple` `additive` edge onto such a species now BECOMES its tendency instead of summing onto a zero. Simulation results do not change — see the Julia section below. |
 | Python | `UnitWarning.path` | Was the string `"unit_validation"` at every site; is now `""` (the document root). | A consumer matching on that literal stops matching. |
 | Python | index-set merge | A model-nested `index_sets` now **merges over** the document-scoped registry instead of being invisible to it. | Different resolution verdicts on pre-0.8.0-shaped documents. |
 | Julia | diagnostic pointers | A scalar `update: {...}` no longer reports a synthetic `/0` segment. | A consumer matching on diagnostic JSON Pointers sees a different path. |
@@ -256,6 +257,36 @@ loaded from a stream therefore kept its subsystem refs as unresolved
 `SubsystemRef`s — and `flatten` **silently skips** an unresolved `SubsystemRef`.
 Same bytes, strictly smaller system, no error. All three entry points now share
 one pipeline; the successor of `load(::IO)` is `load_string(::IO)`.
+
+### A species in no reaction gets no equation (bug fix, silent)
+
+`derive_odes` and `flatten` emitted `D(X, t) = 0` for a species whose net
+stoichiometry is zero in every reaction — the empty sum. Python, Rust, Go and
+TypeScript all omit the equation, so Julia was the 4–1 outlier; the rule is now
+written down as **esm-libraries-spec §4.6.1** and Julia follows it.
+
+Nothing raises. What changes is the **shape of the flattened system**:
+
+| | Before | After |
+|---|---|---|
+| `equations` | carries `D(X, t) = 0` | no entry for `X` |
+| `equation_count` | counts the inert species | does not |
+| equation ORDER | an inert species that a later operator also touches kept the reaction system's position | it takes the operator's position instead |
+| `X` in `state_variables` | yes, DIFFERENTIAL | yes, but in `algebraic_variables` too |
+| a `couple` `additive` edge onto `X` | summed onto the zero (`D(X) = 0 + expr`) | **becomes** the tendency (`D(X) = expr`, §4.7.2) |
+| a `couple` `multiplicative` edge onto `X` | multiplied the zero | raises `couple_multiplicative_no_tendency` — there is no tendency to multiply |
+
+If you pin equation strings, counts, or order, expect those pins to move —
+toward the shared `tests/conformance/flatten` corpus, which the other four
+bindings already matched.
+
+**Simulation is unchanged.** `ModelingToolkit.System(flat)` closes a state that
+no equation mentions with `D(X, t) ~ 0` at the lowering (see
+`ext/mtk_ext/systems.jl`), so such a species stays in `unknowns`, still binds an
+initial condition, and is still held at its initial value in the solution —
+the same trajectory as before, and the same one Python's SciPy backend
+produces. The zero is a backend closure only; it is never written into
+`flat.equations`.
 
 ## TypeScript — `@earthsciml/ast`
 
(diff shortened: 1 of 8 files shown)


ctessum:
It needs to be looked into how this interacts with operator_compose. Because operator compose should still add equation terms for species that don't have any reactions. Does operator_compose match to variables that don't have any pre-existing equations?

PR #235, opened 2026-09-07
ci: stop one failing linter from skipping the rest, and compile the bench targets

### LINKED ISSUE #217: CI: a dirty `cargo fmt` silently skips clippy and four other linters
## What happened

While landing #194 I pushed a branch whose `cargo fmt` was dirty. Fixing that turned the Lint job's `cargo clippy` step from `skipped` to `failure` — and it then reported a genuine dead-code finding on the very first line of code it had ever examined on that branch.

The workflow already suspects this. `conformance-testing.yml:145` carries the comment:

> the clippy step sat behind the rustfmt gate, so it has not actually run in …

## The mechanism

In `.github/workflows/conformance-testing.yml`, the `Lint and format (all languages)` job runs its steps in sequence, and a failing step short-circuits the rest:

```
success  ruff check / ruff format
failure  cargo fmt
skipped  cargo clippy          <-- and every step after it
skipped  Install TypeScript dependencies
skipped  eslint / prettier
skipped  gofmt
skipped  golangci-lint (pinned)
skipped  Version sync (all bindings)
```

So for as long as `cargo fmt` is dirty on `main`, clippy never runs — and neither do eslint/prettier, gofmt, golangci-lint, or the version-sync check. A formatting nit silently disables five real linters.

## Why it matters

These are not equivalent failures. `cargo fmt` is cosmetic and self-fixing; `cargo clippy -- -D warnings` catches correctness and dead-code problems. Ordering them so the cheap cosmetic one can mask the substantive one inverts their value, and the masking is invisible: the job reports one failure, and nothing says four other linters were never consulted.

## Suggested fix

Give each linter its own step with `if: always()` (or `continue-on-error` plus a final aggregation step), so one language's formatting nit cannot suppress another language's linter:

```yaml
- name: cargo fmt
  run: cargo fmt --check
- name: cargo clippy
  if: always()
  run: cargo clippy --all-targets -- -D warnings
- name: eslint / prettier
  if: always()
  …
```

Splitting the job per language would work too, and would make the check names say which linter failed.

## Related

Found while working on #194 (PR #210), which fixes the one clippy finding it exposed in code that PR touches. Sibling issue: the bench targets have drifted out of compiling for a related "nobody actually builds this" reason.

### LINKED ISSUE #218: Rust benches do not compile, and CI never builds them
## What happened

`pkg/earthsci-ast-rs/benches/benchmarks.rs` does not compile, and has not for some time. `cargo check --all-features --benches` reports **7 errors**:

```
error[E0308]: mismatched types
   --> benches/benchmarks.rs:52:35
 52 |     default: Some(1.0),
    |              ---- ^^^ expected `InlineValue`, found floating-point number

error[E0063]: missing field `element_type` in initializer of `ModelVariable`   (line 49)
error[E0063]: missing field `comment` in initializer of `Equation`             (line 66)
error[E0063]: missing field `analyses` in initializer of `earthsci_ast::Model` (line 76)
error[E0063]: missing field `default_units` in initializer of `Species`        (line 137)
error[E0063]: missing fields `analyses`, `tests` and `tolerance`
             in initializer of `ReactionSystem`                                (line 171)
```

Every one is the bench file lagging behind a struct that gained a field: `ModelVariable::element_type`, `Equation::comment`, `Model::analyses`, `Species::default_units`, `ReactionSystem::{tolerance, tests, analyses}`, and `Parameter::default` becoming `InlineValue` instead of `f64`. So the drift accumulated across several unrelated changes, none of which was told about it.

## Why CI never noticed

CI runs `cargo test --all-features` (`conformance-testing.yml`, the `rust-tests` matrix). **`cargo test` does not build bench targets** — that needs `--benches` or `--all-targets`. So nothing in the pipeline ever compiles this file.

It is also invisible to the obvious local command: `cargo check --all-targets` on the default feature set does not reach it either.

## Consequence

The benchmarks cannot be run at all right now — `cargo bench` fails before measuring anything. Any performance regression in the window since the drift started went unmeasured, and anyone reaching for the benches to investigate one has to fix the harness first.

## Suggested fix

Two parts, the second being the one that matters:

1. Repair the six literals (all mechanical: add the missing fields, wrap the `default` in `InlineValue`).
2. **Make CI compile them**, so this cannot recur silently. Cheapest is a check-only step in the existing `rust-tests` job:

   ```yaml
   - name: Build bench targets
     run: cargo check --all-features --benches
   ```

   `cargo check` rather than `cargo bench`: the point is that the code still builds against current types, not to spend CI minutes measuring.

## Related

Found while working on #194 (PR #210), by running `cargo check --all-features --all-targets` to chase a different pre-existing compile error — `tests/performance.rs` had a duplicated `ReactionSystem` field group (`E0062`) that broke the whole `rust-tests` matrix on `main`. That one CI *did* hit, and PR #210 fixes it. This one it never sees.

Sibling issue: `cargo clippy` has been permanently skipped behind a failing `cargo fmt` in the Lint job — the same "nobody actually runs this" shape.

ctessum:
Two defects of the same class — "CI never actually runs this" — both landing in `.github/workflows/conformance-testing.yml`, so they go in one PR.

## #217 — a dirty `cargo fmt` skipped five other linters

The `Lint and format (all languages)` job runs its gates as sequential steps, and a failed step short-circuits the rest. So a rustfmt diff — cosmetic and self-fixing — silently skipped `cargo clippy`, eslint, prettier, gofmt, golangci-lint and the version-sync check. The job reported one failure, and nothing said that three other languages' substantive linters and the cross-binding version check were never consulted.

**Change:** each gate now carries `if: ${{ !cancelled() }}`. The job consults all of them and still fails if any fails, so one push surfaces every finding rather than one finding per round trip.

Per-step conditions rather than splitting the job per language, deliberately: the split job exists to make each linter run *once* rather than once per matrix leg (its header comment says so), and the failed **step** name already identifies which linter went red, which is what the "split for better check names" argument was after.

`!cancelled()` rather than `always()`: this workflow sets `cancel-in-progress: true`, and `always()` would keep linting through a cancellation instead of letting a superseded run stop. The `Setup *` / cache steps keep no condition — linting against a toolchain that failed to install is noise, not signal.

## #218 — nothing ever compiled the bench targets

`cargo test` does not build bench targets (that needs `--benches` / `--all-targets`), so `benches/benchmarks.rs` drifted out of compiling unseen. Reproduced on `origin/main` @ 71e2538 — six errors, every one the bench file lagging a struct that gained a field:

```
benches/benchmarks.rs:52:35: error[E0308]: expected `InlineValue`, found floating-point number
benches/benchmarks.rs:49:17: error[E0063]: missing field `element_type` in ModelVariable
benches/benchmarks.rs:66:28: error[E0063]: missing field `comment` in Equation
benches/benchmarks.rs:76:21: error[E0063]: missing field `analyses` in Model
benches/benchmarks.rs:137:13: error[E0063]: missing field `default_units` in Species
benches/benchmarks.rs:171:5: error[E0063]: missing `analyses`, `tests` and `tolerance` in ReactionSystem
```

**Change (1/2):** the fixture builders named every field of every struct, which is exactly what let each unrelated field addition break them. They now spread `..Default::default()` — the pattern `ModelVariable`'s own docs prescribe. A new OPTIONAL field no longer breaks the benches; a new REQUIRED one on a type with no `Default` (`Reaction` here) still does, which is the distinction worth keeping. The documents built are unchanged: every field dropped from a literal was `None` before and defaults to `None` now, and `EsmFile::default()` sets `esm` to `SCHEMA_VERSION` — the same `"1.0.0"` the literal spelled out.

**Change (2/2), the one that matters:** `rust-tests` now runs `cargo check --all-features --benches`. Check-only — the point is that the benches still build against current types, not to spend CI minutes measuring. Skipped on the MSRV leg for the same reason the test step drops `esio` there (earthsciio → zarrs is MSRV 1.91); benches are dev tooling and make no MSRV claim. It carries `!cancelled()` too, so a red test suite cannot skip it — that would reintroduce #217's shape inside `rust-tests`.

## What the newly-unblocked linters found: nothing

The premise worth correcting: **clippy and the four other gates are not currently being skipped on `main`.** #210 / #221 landed a clean `cargo fmt`, so the last `main` run ([34139452613](https://github.com/EarthSciML/EarthSciAST/actions/runs/34139452613)) already shows all nine lint steps green:

```
✓ ruff check / ruff format   ✓ cargo fmt   ✓ cargo clippy
✓ eslint / prettier   ✓ gofmt   ✓ golangci-lint (pinned)   ✓ Version sync
```

So there is no backlog to absorb here, and this PR fixes the *mechanism* before the next dirty `cargo fmt` re-arms it. No clippy/eslint/gofmt/golangci/version-sync findings were fixed in this PR, because there were none to fix.

One related note, left as a follow-up rather than done here: the comment over the clippy step said `--all-features` does not compile. Both blockers it lists are now fixed — `cargo check --all-features` builds clean (that is what the new bench step relies on). But widening clippy to `--all-features` would lint `esio_provider` and `performance` for the first time and likely surface a fresh backlog under `-D warnings`. That deserves its own change with the findings triaged, so this PR only corrects the comment to say so.

## Also here

`CONTRIBUTING.md`'s Rust workflow snippet said `cargo bench`, which — the bench target being `required-features = ["benchmarks"]` — builds and measures nothing. It now names the feature, as the crate README already does. One line, same defect family: a command in the docs that nobody had actually run.

## Verified

Run in `pkg/earthsci-ast-rs` (rustc 1.96.0; CI pins 1.98.0 for lint):

| command | before | after |
| --- | --- | --- |
| `cargo check --benches --features benchmarks,performance` | 6 errors | clean |
| `cargo check --all-features --benches` (the exact new CI step) | — | clean, exit 0, 5m51s |
| `cargo fmt --all -- --check` | — | clean |
| `cargo clippy --bench benchmarks --features benchmarks,performance` | — | no diagnostics on `benches/` |

And on this PR's own CI run ([34152509153](https://github.com/EarthSciML/EarthSciAST/actions/runs/34152509153)), the lint job is green with **all nine gates executed** under the new conditions — `ruff check / ruff format`, `cargo fmt`, `cargo clippy`, `npm ci`, `eslint / prettier`, `gofmt`, `golangci-lint`, `Version sync` — confirming the `if:` syntax is valid and that nothing new was surfaced. `rust-tests` (with the new bench step) is still running at the time of writing.

Also: `cargo check --benches -v` confirms rustc is actually invoked on `benches/benchmarks.rs` (the target is `required-features = ["benchmarks"]`, so a bare `--all-targets` never reaches it — which is also why CI's clippy step does not cover it), and the workflow YAML parses with every `if:` landing on the intended step.

Not run, per the repo's CI-owns-the-suites split: the conformance harness and the five per-binding test suites. The benches are compiled, not executed — `cargo bench` measurement is out of scope for this fix.

Fixes #217
Fixes #218


ctessum:
CI note: `rust-tests (stable)` shows red, but not for anything in this PR. Every real step passed, including the new one:

```
success  Run Rust tests
success  Build bench targets      <- the new step, on both stable and beta
success  Generate Rust documentation
failure  Post Cache Rust dependencies
```

The only failure is `actions/cache`'s post-run save (`"/usr/bin/tar" failed with exit code 2`), which is infrastructure and pre-existing — the same annotation appears on `main`'s own run [34139452613](https://github.com/EarthSciML/EarthSciAST/actions/runs/34139452613). `rust-tests (beta)` and `rust-tests (1.89.0)` are both green, and `1.89.0` correctly skipped the bench step.

Issue #232, opened 2026-09-07
julia: an array-shaped observed defined by the INDEXED LHS spelling (`y[i] ~ f(…)`) is refused with E_TREEWALK_UNSUPPORTED_SHAPE


ctessum:
## Summary

esm-spec §6.3.1 admits **two** LHS spellings for an equation that defines an
unknown (making it an observed): bare, `y ~ f(…)`, and indexed, `y[i] ~ f(…)`.
For an **array-shaped** observed the Julia tree-walk build accepts only the bare
one. The indexed (aggregate-LHS) spelling reaches `_partition_variables`, which
requires an array-shaped observed to be a geometry ring, and throws:

```
simulation failed: E_TREEWALK_UNSUPPORTED_SHAPE: Column.w
```

Rust and Python both run the same document and produce the right numbers, so
this is a cross-binding divergence, not an under-specified document.

Found incidentally while drafting the conformance fixture for #219 (see #229).
I have **not** surveyed how often the indexed spelling is used for an array
observed in real documents, and I did not attempt a fix — I rewrote the fixture
to avoid the intermediate observed rather than widen that PR.

## Minimal reproducer

The same array-shaped observed `w = 2·u`, defined two ways. Only the LHS differs.

```json
{
  "esm": "1.0.0",
  "metadata": { "name": "ArrayObservedAggregateLhs", "authors": ["repro"] },
  "index_sets": { "lev": { "kind": "interval", "size": 4 } },
  "models": {
    "Column": {
      "tolerance": { "abs": 1e-12, "rel": 1e-09 },
      "variables": {
        "w": { "type": "unknown", "units": "1", "shape": ["lev"] },
        "u": { "type": "unknown", "units": "1", "default": 1.0, "shape": ["lev"] }
      },
      "equations": [
        {
          "_comment": "w[k] ~ 2*u[k] -- the INDEXED (aggregate) LHS spelling.",
          "lhs": {
            "op": "aggregate", "args": [], "output_idx": ["k"],
            "ranges": { "k": { "from": "lev" } },
            "expr": { "op": "index", "args": ["w", "k"] }
          },
          "rhs": {
            "op": "aggregate", "args": [], "output_idx": ["k"],
            "ranges": { "k": { "from": "lev" } },
            "expr": { "op": "*", "args": [2.0, { "op": "index", "args": ["u", "k"] }] }
          }
        },
        {
          "lhs": {
[...]

Issue #224, opened 2026-09-07
`standard-conformance-testing` exceeds its 45-minute timeout, including on main, and gates `conformance-results` behind it


ctessum:
## Summary

The cross-language conformance harness routinely exceeds `timeout-minutes: 45` in `.github/workflows/conformance-testing.yml`. Observed on PR #197 at **45m16s**, and **the last several runs on `main` itself are all `cancelled`**.

## Why it matters beyond the job itself

The workflow has a one-way dependency chain:

- `standard-conformance-testing` — `needs: [julia-tests, typescript-tests, python-tests, rust-tests, go-tests]` (line 470)
- `conformance-results` — `needs: [standard-conformance-testing]` (line 563)

So the timeout does not just lose the conformance signal: `conformance-results` then fails on a missing artifact, and the PR shows a red check that reads as a conformance regression when nothing conformance-related happened. The same cascade fires whenever any one language job fails for an unrelated reason.

## Effect

Every PR inherits a red or missing conformance signal regardless of its content, which trains reviewers to ignore the check — the failure mode where a real regression gets waved through.

## Options worth considering

- Raise `timeout-minutes` (treats the symptom; the harness re-runs every language's full suite before comparing, so it will keep growing).
- Split the harness so bindings run in parallel rather than sequentially.
- Cache or reuse the per-language suite runs the individual jobs already performed, instead of re-running them inside the harness.
- Make `conformance-results` degrade gracefully when its upstream is skipped or cancelled, so a missing artifact reports as "not run" rather than "failed".

ctessum:
## Data: across the last 30 conformance workflow runs

| `standard-conformance-testing` outcome | count | interpretation |
|---|---|---|
| `cancelled`, ~0s duration | 16 | superseded by a newer push (concurrency group) — normal, not a problem |
| `skipped`, ~0s duration | 10 | an upstream language job failed, so `needs` was never satisfied |
| `cancelled`, **45.2–45.4 min** | 4 | genuine timeout against `timeout-minutes: 45` |

**The job executed 4 times out of 30, and all four ended at the wall** (45.3, 45.3, 45.4, 45.2 min).

Two things follow that are worth stating precisely, because the raw `cancelled` count is misleading:

1. Most cancellations are *not* timeouts. Filtering the run list by conclusion overstates the problem; only job duration distinguishes a superseded run from a timed-out one.
[...]

PR #237, opened 2026-09-07
fix(python): a declared `shape` decides the pathway, and §9.6.3 is a real pre-evaluation walk

### LINKED ISSUE #231: python: a bare whole-array `D(state)` over a declared `shape` routes to the SCALAR pathway, so the state has no cells
## Summary

Python's `_choose_pathway` decides between the scalar (SymPy) and array (NumPy)
runtimes purely from **equation content** — it never consults a variable's
declared `shape`. So a document whose only array-ness is a declared `shape`, with
a bare whole-array derivative and no `aggregate`/`arrayop` node anywhere, routes
to the **scalar** pathway. The shaped state then has no cells, and an assertion
with `coords` cannot find it.

Found incidentally while fixing #219 (see #229). I have **not** surveyed how often
this spelling occurs in real documents, and I did not attempt a fix — it is on a
different axis from #219 and I did not want to widen that PR.

## Minimal reproducer — no parameters at all

This matters because #219 was about a shaped *parameter*. This document has no
parameters whatsoever, so the two defects are independent:

```json
{
  "esm": "1.0.0",
  "metadata": { "name": "BareShapedNoParams", "authors": ["repro"] },
  "index_sets": { "lev": { "kind": "interval", "size": 4 } },
  "models": {
    "Column": {
      "variables": {
        "theta": { "type": "unknown", "units": "K", "default": 1.0, "shape": ["lev"] }
      },
      "equations": [
        { "lhs": { "op": "D", "args": ["theta"], "wrt": "t" }, "rhs": 1.0 }
      ],
      "tests": [
        {
          "id": "theta_ramps",
          "time_span": { "start": 0.0, "end": 1.0 },
          "assertions": [
            { "variable": "theta", "time": 1.0, "coords": { "lev": 2 }, "expected": 2.0 }
          ]
        }
      ]
    }
  }
}
```

```
>>> esm_problem(f, (0.0, 1.0)).pathway
'scalar'
>>> run_pde_tests(f)
passed=False  msg=assertion evaluation failed: array state 'theta' has no cells in
              var_map, and no array observed of that name is exposed by the build
              or evaluable at the assertion time
```

Rewriting the *same* equation in the `aggregate` spelling — `aggregate{k}(D(theta[k])) ~ aggregate{k}(1.0)`,
identical semantics — routes differently and works:

```
>>> esm_problem(f_agg, (0.0, 1.0)).pathway
'array'
>>> run_pde_tests(f_agg)
passed=True  actual=2.0000000000000004
```

A related symptom appears when a shaped **parameter** in the same bare-spelling
document carries inline array data: the scalar pathway detects the value it
cannot bind and says so explicitly, which is where the routing gap is visible in
the error text rather than in a missing cell:

```
solve failed: 'Column.ramp' carries inline ARRAY data (esm-spec §6.3 / §6.6.2),
which this pathway cannot bind: a nested-array `default` / `parameter_overrides`
/ `initial_conditions` value requires the array simulation pathway.
```

## Where it is

`pkg/earthsci-ast-py/src/earthsci_ast/problem.py`, `_choose_pathway`:

```python
    if discrete_providers:
        return "discrete_providers"
    if merged or gated:
        return "array"
    if flat.loader_fields:
        return "loaders"
    if any(_has_array_op(eq.lhs) or _has_array_op(eq.rhs) for eq in flat.equations):
        return "array"
    return "scalar"
```

Every arm is about providers, loaders, or an array **op**. A variable's declared
`shape` — which esm-spec §6.3 makes the authoritative statement of array-ness,
and which §11 already treats as authoritative over usage inference in the array
build — never enters the decision.

## Observed vs. what the spec says

esm-spec §6.3 defines `shape` as "the ordered list of index-set names the
variable is arrayed over"; nothing there makes the array-ness conditional on how
the defining equation happens to be spelled. Rust runs the bare form over the
same declared shape and produces the right numbers (it is pinned in
`pkg/earthsci-ast-rs/tests/shaped_parameter_broadcast_conformance.rs`,
`bare_whole_array_derivative_over_a_broadcast_parameter`), so this is a
Python-side routing gap rather than an under-specified document.

I have not checked whether the fix is as simple as adding a
"any variable declares a non-empty resolvable `shape`" arm; that would change
pathway selection for documents that currently take the scalar route, so it
wants its own look.

## Why the corpus does not catch this

The conformance fixture added in #229
(`tests/conformance/shaped_parameter_broadcast/`) deliberately uses the
`aggregate` spelling, because that is the one every executing binding routes to
its array runtime — a shared fixture cannot pin the broadcast contract and this
routing gap at the same time. The Rust runner in that PR covers the bare form
separately, precisely because Rust's whole-array lowering exercises a gather the
`aggregate` spelling does not. So nothing in the shared corpus currently drives
the bare spelling through Python, and a fix here would want a fixture that does.

ctessum:
Fixes #231

Three Python changes on array-shaped variables. **Part 1** is #231's pathway
routing — the reason this PR exists. **Parts 2 and 3** followed from reviewing
it: routing on a declared `shape` exposed that the layout pass and the §9.6.3
gate did not read the same authority.

> **Scope note.** An earlier revision of this branch also carried #232's Python
> half (`57b72acad`, the indexed-LHS array observed). That work has been split
> out into #276 and reverted here, so this PR is scoped to declared-shape
> routing and its two consequences. #276 is the sibling of #250 (the Julia half).

# Part 1 — a declared `shape` must decide the pathway (#231)

## The bug

Python's `_choose_pathway` decided between the scalar (SymPy) and array (NumPy)
runtimes purely from **equation content** — an injected array, a `loader_fields`
seam, or an `index` / `aggregate` / `arrayop` node somewhere in an equation. A
variable's declared `shape` never entered the decision.

So a document whose only array-ness is a declared `shape` — a bare whole-array
`D(theta) ~ 1` over `"shape": ["lev"]`, with no array op anywhere — routed to
the **scalar** pathway, where the shaped state was laid out as a single scalar
slot and a `coords` assertion could not find it:

```
>>> esm_problem(f, (0.0, 1.0)).pathway
'scalar'
>>> run_pde_tests(f)
passed=False  msg=assertion evaluation failed: array state 'theta' has no cells in var_map…
```

The *same* equation in the `aggregate` spelling routed to the array pathway and
worked. The spelling, not the model, was deciding the answer. Rust runs the bare
form over the same declared shape and produces the right numbers, so this was a
Python routing gap, not an under-specified document.

## The change

`pkg/earthsci-ast-py/src/earthsci_ast/problem.py` — one new arm in
`_choose_pathway`, backed by `_declares_resolvable_shape`:

```python
    if flat.loader_fields:
        return "loaders"
    if _declares_resolvable_shape(flat):
        return "array"
    if any(_has_array_op(eq.lhs) or _has_array_op(eq.rhs) for eq in flat.equations):
        return "array"
    return "scalar"
```

esm-spec §6.3 makes `shape` — "the ordered list of index-set names the variable
is arrayed over" — the authoritative statement of array-ness, and §11 already
treats it as authoritative over usage inference *inside* the array build:
`_build_numpy_rhs` resolves declared shapes against the `index_sets` registry
before it lays out the state vector. The routing now reads the same authority,
through the **same** `_resolve_index_set_shape` resolver, and inherits the same
fallback:

- A shape counts only when **every axis resolves** to a concrete extent. An
  unresolvable shape (an axis naming no registry entry, a `derived` set whose
  extent value-invention has not materialized) is exactly the case where the
  array build would fall back to usage inference and infer the same scalar — so
  routing on it would change the engine without changing the answer.
- The arm sits **after** `loader_fields`, so the cadence-segmented pathway is
  untouched. It already builds through `_build_numpy_rhs`, so it already honours
  declared shapes.
- It consults states, parameters **and** observeds: every §6.3 variable role
  carries `shape`, and a shaped parameter with inline array data is array-valued
  whatever its consumers look like.

This also closes the related symptom the issue records: a shaped parameter
carrying §6.3 inline array data on the same bare spelling failed with
`'Column.ramp' carries inline ARRAY data … which this pathway cannot bind`. It
now routes to the pathway that binds it, and each cell integrates its own rate
(pinned by a test).

## Corpus impact

Swept every loadable document under `tests/valid/`, `tests/conformance/` and
`tests/fixtures/` (407 candidate files, 259 that load *and* flatten), computing the pathway
with the routing arm off and on in a single pass:

| pathway | before | after |
|---|---|---|
| array | 146 | **162** |
| scalar | 94 | **78** |
| loaders | 19 | 19 |

**16 documents move from `scalar` to `array`.** Nothing else moves. All 16
declare a resolvable `shape` on a state, so all 16 were laying out a multi-cell
field as a single scalar slot — the move is a correctness fix for each of them,
not a regression:

- `tests/valid/coordinates_registry.esm`
- `tests/valid/derivative_trailing_boundary_operands.esm`
- `tests/conformance/classification/fixtures/subsystem_scope.esm`
- `tests/conformance/output_derivation/fixtures/gridded.esm`
- `tests/conformance/expression_templates/{constrained_match_scope,
  import_where_rename_two_instances, opacity_priority_shadowing,
  per_variable_scheme_literal_args, two_div_two_meshes}/{fixture,expanded,emitted}.esm`
  (12 files)

**All 16 still build.** No Python test builds or simulates any of them today
(the four test files that reference them — `test_expression_templates.py`,
`test_out_of_line_templates.py`, `test_template_imports.py`,
`test_roundtrip_against_original.py` — exercise load / expand / round-trip only),
and all four still pass.

Seven further scalar-pathway documents declare a shape that resolves against no
`index_sets` entry (`["lon","lat"]` / `["x","y"]` with no such registry
entries — `expression_templates/inject_*`, `tests/fixtures/arrayed_vars/*`) and
are deliberately **left alone** by the resolvability guard.

### One behaviour change worth a reviewer's eye

`constrained_match_scope/{fixture,expanded}.esm` previously **failed to build**
with `unlowered_operator: rewrite-target op 'div' …` and now build and solve.
The cause is a pre-existing asymmetry between the two pathways, not something
this PR introduces: those documents carry a surviving unlowered `div` in a
**dead** observed (`div_cell`, which nothing consumes — that is the point of the
match-scoping fixture). The scalar pathway lambdifies every observed eagerly and
so trips over the dead one; the array pathway evaluates observeds lazily, so a
dead observed is never reached and the `unlowered_operator` gate — which
esm-spec §9.6.3/§9.6.8 describes as a *pre-evaluation* gate — never fires. This
PR only moves those two fixtures across that pre-existing line. Nothing asserts
the old build failure.

# Part 2 — the declared shape must reach the LAYOUT, not just the routing (`0acb0c80f`)

`_declares_resolvable_shape` consults `state_variables`, `parameters` **and**
`observed_variables` when choosing the pathway, but `_build_numpy_rhs`'s own
declared-shape resolution looped only over `flat.state_variables`. An observed's
declared shape could therefore flip the engine without the build using that
shape for anything.

`_build_numpy_rhs` now resolves declared shapes for observeds too, so routing
and layout read one authority. `test_the_shape_arm_reads_states_parameters_and_observeds`
gained the observed-only case its name already claimed.

**Corpus impact: none.** Zero documents change build state.

# Part 3 — §9.6.3 constraint 6 is a WALK, not a reachability test (`e0a67cee2`)

esm-spec §9.6.3 constraint 6 says the gate is a **pre-evaluation walk of the
component's expression trees**: *"before a component is EVALUATED or COMPILED
for simulation, its expression trees are walked; any node whose `op` is not in
the evaluable-core set … is rejected."*

Python implemented no such walk. It raised `unlowered_operator` **reactively at
evaluation**, so the scalar pathway (which lambdifies every observed eagerly)
tripped over an unlowered op in a *dead* observed, while the array pathway
(lazy observeds) never reached one. The array pathway was therefore
under-enforcing a normative gate.

Part 1 did not create that asymmetry, but it moves two fixtures across the line,
which is how it surfaced. `_assert_no_unlowered_operator` now walks equations
and `ic` RHSs at the `esm_problem` front door.

## Corpus impact — measured on the merged tree, and it is not what an earlier revision said

Re-run against `origin/main` after merging 111 commits, over 312 documents:

| | before | after |
|---|---|---|
| scalar | 85 | **71** |
| array | 143 | **157** |
| `UnreachableSpatialOperatorError` | 5 | **5** |

**14 documents move scalar→array. No document changes build state.**

An earlier revision of this description said the walk takes
`constrained_match_scope/{fixture,expanded}.esm` from building to rejected. On the
merged tree that is **no longer true, and the walk's role has inverted**. Those two
fixtures are *already* rejected on `main`: without the Part 1 shape arm they route
scalar, where eager lambdify catches the unlowered `div` at `expression.py:492`.

Isolating the walk on the merged tree, by disabling only it:

| configuration | `constrained_match_scope/fixture.esm` |
|---|---|
| `origin/main` | REJECTED (scalar, eager) |
| Part 1 shape arm, **walk disabled** | **BUILDS (array)** ← regression |
| this PR (shape arm + walk) | REJECTED |

So the walk does not *cause* those rejections any more — it **stops Part 1 from
regressing them**. Part 1 alone would route both to the lazy array pathway, where
nothing would ever reach the unlowered node. The walk remains load-bearing:
disabling it fails 2 of the gate tests.

# Correction to an earlier revision of this description

A previous revision claimed "All 16 still build" of the documents Part 1 moves
scalar->array. That claim was wrong, and it is also stale: on the merged tree the
mover set is **14**, not 16, and its membership changed. `constrained_match_scope`
and `opacity_negative` are no longer in it; two fixtures that arrived from `main`
are (`toplevel_ref_index_sets/toplevel_ref_metaparameter_axis{,_lib}.esm`).

The accurate statement is the table above: 14 movers, zero documents changing
build state.

# Verified

- Full Python suite, on the merged tree: **3320 passed, 16 skipped, 3 xfailed, 0 failed**. The three
  excluded files (`test_esio_provider.py`, `test_loader_ingest_and_select.py`,
  `test_loader_unit_conversion_conformance.py`) fail only from a stale local
  `earthsciio` lacking `DataSource`; CI's four `python-tests` legs are green.
- PDE-simulation conformance, Python leg: **8/8 OK** against the Julia golden
  and analytic anchors.
- Full-pipeline PDE conformance, Python leg: **2/2 OK**.
- `ruff check src/ tests/ ../../scripts/` and `ruff format --check`: clean.
- The 13 new tests genuinely fail without the changes: 6 shape tests against
  main's source, and 2 gate tests with only the walk disabled. Note `pyproject.toml` sets `pythonpath = ["src"]`, which silently
  overrides `PYTHONPATH` under pytest — a baseline run needs `-o pythonpath=…`.

This PR now carries **zero net Rust change**. `d16bc91ad` had wrapped a
conformance test's tolerances in `Some(...)` because main did not compile at the
time; #252 landed the byte-identical fix, so the merge took main's version.

# Known-unrelated CI

`standard-conformance-testing` is red here and on **every** branch including
main — its Julia leg alone exceeds the job's 45-minute timeout. Tracked as #224.
`conformance-results` fails downstream of it. Do not block this PR on either.

# Follow-ups

- The §9.6.3 walk landed in Python only. Julia has the same gap in a worse form
  (`_fold_elementwise_array_observeds` deletes dead observeds before any walk
  could see them); Go and TypeScript both ship public evaluators, so neither is
  exempt. Worth a tracking issue.
- `_declares_resolvable_shape` reads observeds and parameters, but the build
  honours only states' declared shapes for layout beyond Part 2's change.


ctessum:
## Three Python divergences for this PR, with reproducers — from #250

**#250** (Julia, issue #232) fixes the Julia half of the indexed observed-LHS
spelling (`y[i] ~ f(…)`, esm-spec §6.3.1) and adds a shared conformance
category, `pde_inline_observed_indexed_lhs`. **Python is deliberately red
against that category**, kept in `bindings_required` rather than
`scope_excluded` — it has a real runner, so the category states the §6.3.1
contract and lets Python be red until it holds.

The coordinator asked me to hand these to you rather than fix Python myself.
Three complete, self-contained reproducer documents are in the body of #250,
under "Reproducers for the Python divergences (for #237)" — copy them out and
run them directly.

### The three

1. **An indexed-LHS array observed is not readable by an assertion.** Asserting
   it at any time returns `0.0` instead of its field value. It *is* reaching
   the dynamics (the driven state integrates correctly) — only the assertion
   read is empty. Julia returns `14.778112197861757` where Python returns
   `0.0`.

2. **An indexed LHS with a PER-CELL rhs is silently dropped.**
   `aggregate{k}(w[k]) ~ 2*u[k]`, with no `aggregate` on the right (the `k` is
   bound by the LHS frame), raises

   ```
   RuntimeWarning: solve: unrecognized algebraic equation with LHS
   ExprNode(op='aggregate', … expr=ExprNode(op='index', args=['Column.w','k'])…)
   was not applied to the ODE RHS; any state it constrains stays frozen at its
   initial value
     (simulation_array.py:2671, in _apply_equation_to_dy)
   ```

   and the state does stay frozen. This is the shape that yields a plausible
   wrong number rather than an error.

3. **An indexed-LHS observed feeding a WHOLE-ARRAY derivative is dropped the
   same way.** `D(u) ~ wf` leaves `u` at its initial value; respelling only the
   derivative as `aggregate{k}(D(u[k])) ~ aggregate{k}(wf[k])` makes it
   integrate. The #250 reproducer for this one is the committed conformance
   fixture with *only* the two derivative equations changed, so diffing it
   against
   `tests/conformance/pde_inline_observed_indexed_lhs/fixtures/observed_indexed_lhs.esm`
   isolates the derivative spelling exactly.

### I measured against this branch rather than guessing

Divergence 3 reads exactly like this PR's title, so I expected it to be already
fixed here. **It is not.** I extracted `pkg/earthsci-ast-py/src` from this PR at
head `67504523` with `git archive`, put it on `PYTHONPATH`, and ran all three:

```
divergence 1   1 pass  2 pass  3 FAILED actual=0.0          identical to main
divergence 2   1 pass  2 FAILED actual=1.0                  identical to main
divergence 3   1,2,3,5,6 FAILED actual=0.0   4 pass         identical to main

#250's conformance fixture on this branch: 7/13 — the same 7 as on main
```

So all three are still open work here. Worth a look at whether they are one
root cause or three: 2 and 3 share the "equation not applied to the ODE RHS"
symptom, while 1 is on the assertion-read path and may be independent.

### The check to re-run

`tests/conformance/pde_inline_observed_indexed_lhs/` goes green for Python at
13/13. Today it is 7/13: the seven state assertions pass, and the six non-zero
observed assertions (`wf` and `ws`, both indexed-LHS array observeds) all return
`0.0`. Runner:
`pkg/earthsci-ast-py/tests/test_pde_inline_observed_indexed_lhs_conformance.py`.
Julia and Rust both pass 13/13.

Context for the category, the three divergences, and their expected-failure
status: `tests/conformance/pde_inline_observed_indexed_lhs/README.md` and
CONFORMANCE_SPEC §5.30.1.

ctessum:
## Re-measured at `57b72acad` — all three closed, and the fixture is now 14 assertions

Following up my earlier comment, which measured `67504523` and found none of the
three fixed. I re-ran the same way against **`57b72acad`** — your
`pkg/earthsci-ast-py/src` extracted with `git archive`, put on `PYTHONPATH`,
nothing else changed:

```
divergence 1   1 pass  2 pass  3 pass actual=14.778112198680008
divergence 2   1 pass  2 pass actual=7.389056099340004
divergence 3   all 6 pass
```

All three closed. Independent confirmation of your own result, from the other
side.

### Your root cause is the same one #250 fixes in Julia

`flatten._collect_model` reading observed-ness from
`classification.inlined_unknowns` — the strict `y ~ f(…)` set §6.3.1 sanctions
for *inlining specifically* — and using it as the classification, is exactly the
mistake #250 fixes on the Julia side, where the tree-walk build's owner buckets
made it syntactically via `eq.lhs isa VarExpr`. §6.3.1's "it does not narrow the
partition" is the sentence both bindings walked past. Two bindings, one wrong
reading, diagnosed from opposite ends — which is about as good as
cross-validation gets.

### The fixture has changed under you — please re-run

Two things moved since you measured 13/13:

1. **It is now 14 assertions, not 13.** You and I independently flagged that
   `ws@t=0` is a false pass (it is genuinely zero, so a binding that always
   answers zero passes it). I kept it — it pins the state-dependent observed at
   the trajectory start — and added **assertion 8, `ws` at `t = 0.5`, expected
   `9.0`**, as its discriminating partner. `ws` is now sampled at three distinct
   times and only one of them can be passed by accident. The golden is re-minted.
2. **CONFORMANCE_SPEC §5.30 became §5.32.** `main` claimed §5.30 and §5.31 while
   both our PRs were open; #250 renumbered on rebase. If you cite it anywhere,
   it is §5.32 / §5.32.1 now.

I verified the 14-assertion version against `57b72acad`: **14 / 14**, including
the new assertion 8 at `9.000000000000007`. So nothing is broken by the change —
but your branch's copy of the fixture, manifest and golden is stale, and re-running
after a rebase onto #250 (or onto `main` once #250 lands) is worth doing.

### Where it stands

`main` 7/14 · `67504523` 7/14 · **`57b72acad` 14/14**. Julia and Rust: 14/14.

#250 documents Python as "red on `main`, fixed by #237, pending merge" rather
than as a standing divergence — in the fixture README, CONFORMANCE_SPEC §5.32.1,
the manifest description, the compliance-matrix row, and the Python adapter
docstring. Nothing there needs editing when you merge; the tag
`red-on-main-until-pr-237` and that prose just go stale and should be dropped.
Deliberately **no `xfail`** — it would flip to an unexpected-pass failure the
moment you land.

PR #238, opened 2026-09-07
feat(units): add inHg, exactly 25.4 mmHg

ctessum-claude:
## What

Adds `inHg` to the pressure registry, at exactly 25.4 × the existing `mmHg` entry.

```rust
units.insert("mmHg".to_string(), pascal_scaled(&units, 133.322_387_415));
units.insert("inHg".to_string(), pascal_scaled(&units, 3_386.388_640_341));
```

Plus the `esm-spec.md` §1110 pressure row and a test.

## Why

The registry had `mmHg` and no route to inches of mercury. That gap is not cosmetic, because there is **no escape hatch**: a unit string carries no numeric scale factor, so `25.4 mmHg` is not spellable either. Measured against the pinned toolchain, all of these are refused — `inHg`, `in`, `inch`, `in_i'Hg` (the UCUM spelling), `25.4 mmHg`, `25.4*mmHg`, `mmHg*25.4`, `inch*mmHg`.

So a column holding inches of mercury had exactly two options, and both defeat the point of the field:

- declare a **wrong** unit (`mmHg`, off by 25.4), or
- declare **no** unit, which turns the one arithmetic step a dimensional check could catch into an unchecked magic number.

This surfaced porting EPA MOVES, whose `County.barometricPressure` is in inHg and whose every consumer multiplies by 3.38639 to reach kilopascals. The downstream document currently declares `units: "1"` on three pressure variables with a paragraph of apology in each `description`.

## The constant

3386.388640341 Pa is the NIST SP 811 conventional value, and it is exactly 25.4 × the `mmHg` entry already in the registry — the two agree **bit-for-bit** in f64, so the test asserts equality rather than a tolerance. Deriving it from the neighbouring entry rather than typing an independent literal is deliberate: it cannot drift away from `mmHg`.

## Verification

`cargo test --lib units::` → 53 passed, including the new `test_inches_of_mercury`.
`cargo test --test units` → 17 passed, including `every_unit_string_in_the_valid_corpus_parses`.

End-to-end A/B on a document declaring `units: "inHg"`:

| build | `esm validate` |
|---|---|
| `main` | `✗ Validation failed — Unit string 'inHg' is not a recognised unit` |
| this branch | `✓ Validation passed`, and the document's assertion passes 1 / 0 / 0 |

`inHg` is not added to `CORPUS_VALID_UNITS` — that list is the set of units appearing in `tests/valid/**`, and no fixture there uses it.

## Scope

This is the **narrow** fix. It does not address the general shape of the problem — the registry is a closed list, and a unit string cannot carry a scale factor — which is what would let the next dataset in an unlisted unit declare itself without a spec change. A numeric scale factor in unit strings (so `25.4 mmHg` denotes what it says) would generalise past this case and past pressure; imperial length (`in` — note `ft` is already present) would let `inch` compose if a scale factor existed. Both are larger conversations and neither is attempted here.

https://claude.ai/code/session_012EVRsbWRqAJ1QpULBh2BRo

### DIFF
--- a/esm-spec.md
+++ b/esm-spec.md
@@ -1107,7 +1107,7 @@ A binding MUST NOT introduce a ninth axis, and MUST NOT map a registry symbol on
 | Volume | `L` `l` `mL` |
 | Amount | `kmol` `mmol` `umol` `nmol` `M` |
 | Derived | `Hz` `N` `Pa` `J` `kJ` `cal` `kcal` `W` `kW` `MW` |
-| Pressure | `atm` `uatm` `bar` `hPa` `kPa` `mbar` `Torr` `mmHg` `psi` |
+| Pressure | `atm` `uatm` `bar` `hPa` `kPa` `mbar` `Torr` `mmHg` `inHg` `psi` |
 | Energy | `erg` `BTU` `Wh` `kWh` |
 | Electromagnetic | `C` `V` `Ohm` `F` `T` |
 | Temperature / angle | `degC` `degF` `deg` |
--- a/pkg/earthsci-ast-rs/src/units.rs
+++ b/pkg/earthsci-ast-rs/src/units.rs
@@ -1823,6 +1823,11 @@ fn build_base_units() -> HashMap<String, Unit> {
     units.insert("bar".to_string(), pascal_scaled(&units, 1e5));
     units.insert("Torr".to_string(), pascal_scaled(&units, 133.322_368_421));
     units.insert("mmHg".to_string(), pascal_scaled(&units, 133.322_387_415));
+    // Inch of mercury — exactly 25.4 mmHg, the conventional value (NIST SP 811).
+    // US barometric datasets store pressure in inHg; without this entry such a
+    // column has no honest declaration, because a unit string carries no numeric
+    // scale factor, so `25.4 mmHg` cannot be spelled either.
+    units.insert("inHg".to_string(), pascal_scaled(&units, 3_386.388_640_341));
     units.insert("psi".to_string(), pascal_scaled(&units, 6_894.757_293_168));
 
     // Dobson unit: areal number density of ozone molecules.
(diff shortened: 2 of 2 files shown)


ctessum-claude:
Confirming downstream. `inHg` at exactly 25.4 mmHg is what we needed, and the
repro we carried for it went green on the rebuild.

`County.barometricPressure` in MOVES is in inches of mercury. Four documents in
our port carried `units: "1"` plus a paragraph of prose apologising for it; they
now carry `units: "inHg"` and the prose is gone.

**Declaring the unit found a real inconsistency the moment it was honest.** Two
fixtures gather the county pressure onto a run relation, and that downstream
variable was still declared dimensionless:

```
✗ /models/ProcessNoxSpeciation/equations/31: Left-hand side has units of
    dimensionless but right-hand side has units of length^-1*mass*time^-2
✗ /models/ProcessNoxSpeciation/variables/run_barometricPressure: Declared units
    of dimensionless do not match the expression's units of length^-1*mass*time^-2
```

Good diagnostic — it names the equation, the variable, and both dimensions.

**One observation that is not a request, in case it is useful when someone
revisits the units checker.** We probed what the declaration actually binds, and
it binds on equations that carry a unit-bearing right-hand side, not on the
whole chain:

| change | verdict |
|---|---|
| a *gathered* pressure variable declared `m` (wrong dimension) | refused, naming equation 11 and both dimensions |
| the *ingested* pressure variable declared `m` (wrong dimension, RHS is data with no units) | accepted |
| all three declared `mmHg` (right dimension, wrong by 25.4×) | accepted |

The second and third are both reasonable — a data-fed variable has no RHS units
to disagree with, and a scale error is not a dimension error — but together they
mean the one step our documentation claimed a unit check would catch, `inHg ×
3.38639 → kPa`, is not caught by any declaration, since 3.38639 is a bare
number. We have corrected our own finding to say so. The general fix we
originally asked for — a numeric scale factor in a unit string, so `25.4 mmHg`
denotes what it says — is what would change that, and it is deliberately not
part of #238.

Issue #206, opened 2026-09-06
Rust `esm test` executes no `reaction_systems[].tests`, and an observed defined as `D(x, t)` of an ODE state evaluates to 0 — species tendencies cannot be asserted


ctessum-claude:
## Context

EqWeFiC's first chemistry component (WRF-Chem's RADM2 mechanism as an esm `reaction_systems` entry with a sibling model holding the rate coefficients) exposed two gaps that together make species tendencies untestable.

## 1. `reaction_systems[...].tests` are not executed by the Rust CLI

`esm test` reports "no inline tests found" for a document whose only tests are on a reaction system (repro `probe_rs_only.esm`; the same happens for EarthSciModels' `superfast.esm`). esm-spec §7.2 admits `tests` on reaction systems. Either the Rust runner should discover and run them, or the spec should say they are model-only.

## 2. An observed defined as `D(x, t)` of an ODE state evaluates to 0

An observed `dxdt = D(x, t)` where `x` is an ODE state (own state, or a sibling system's species reached by scoped reference) evaluates to exactly 0 in the Rust runner (repros `probe_dobs_own.esm`, `probe_rs_scoped.esm`). The expected value is the right-hand side of the state's ODE. With this, the standard EqWeFiC test shape (assert tendencies at t = 0) cannot be applied to reaction systems even from a wrapper model.

## 3. (binding divergence) bare `parameter_overrides` names

A bare override name is resolved per component in Rust but document-wide in Python: a model parameter `M` and a constant species `M` in a sibling reaction system passed in Rust and raised `AmbiguousParameterError` in Python. Related to the override-resolution rules in PR #179.

Also observed: document-level `expression_templates` are not visible to models (`probe_units.esm`); libraries must be imported — fine if intended, worth stating in the spec.

Repros are in a README-indexed directory; happy to attach them. Rust CLI built from the `inline-test-array-observeds` branch (PR #177).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_015LxnQcZYdcu9qGuUkncwvc

PR #245, opened 2026-09-07
fix(rust): gate the scalar interpreter on unevaluable operators

### LINKED ISSUE #220: Rust scalar interpreter returns NaN for an unevaluable operator, where the array interpreter reports a diagnostic
## What happens

The two Rust evaluators disagree about what to do with an operator that has no evaluation rule.

The **array** interpreter gates. `simulate_array/compile.rs` stage (0) (`reject_unlowered_spatial_ops` → `check_evaluable`) refuses the whole model up front, and the author gets `unevaluable_operator: operator '<op>' … must be eliminated by an earlier pipeline stage`. That gate exists precisely because `eval_op`'s backstop is `unreachable!`, and the note on `CompileError::UnevaluableOperatorError` spells out the reasoning: *"It is reported rather than evaluated to a NaN sentinel: a silent NaN is indistinguishable from a legitimate numerical result and would propagate into the solution."* `tests/evaluable_core_gate.rs` pins that for all nine gated ops.

The **scalar** interpreter (`Compiled::from_file` / `simulate/`) has no such gate. Feed it the same unevaluable operator and it produces `NaN`, silently, and the NaN propagates into the solution exactly as that comment says it must not.

## How it was found

While fixing #188 (the §9.5.3 `table_lookup` lowering was never applied on the evaluation path). The minimal repro there — a scalar document whose observed is `y = table_lookup(...)` — did **not** report `unevaluable_operator`. It reported:

```
FAIL  TableLookupObserved/table_lookup_matches_hand_lowered[1] (y@t=0.0)
      actual=NaN expected=25 (rtol=0.000001, atol=0)
```

The array-runtime documents in the same issue *did* get the named diagnostic. Same defect, two different author experiences: one names the pipeline stage at fault, the other hands back a number.

#188 removed the specific cause (`table_lookup` is now lowered before either evaluator sees it). The **class** is untouched: any op in the evaluable-core set that reaches the scalar interpreter without a rule still NaNs rather than reporting. That is the same silent-NaN failure mode the array gate was written to close, on the other path.

## What to do

Give the scalar interpreter the same stage-(0) treatment the array one has: refuse a model containing an op it has no rule for, with `unevaluable_operator` naming the op, instead of evaluating it to `NaN`. Then extend `tests/evaluable_core_gate.rs` — which today asserts "no document both VALIDATES and PANICS" for the array runtime — to assert "no document both VALIDATES and silently returns NaN" for the scalar one, over the same op list.

Worth checking whether the other four bindings' scalar paths have the same hole while this is being done.

## Not urgent, but not cosmetic

Nothing currently reaches it — `table_lookup` was the live route and it is closed. It is filed because the invariant the array gate encodes ("an unevaluable op is a diagnostic, never a number") is a property of the *format*, and it is currently enforced on one of two evaluators.

ctessum:
Fixes #220

## The bug

The crate has two evaluators and they disagreed about an operator with no evaluation rule.

The **array** interpreter gates. `simulate_array/compile.rs` stage (0) (`reject_unlowered_spatial_ops` → `check_evaluable`) refuses the model up front with `unevaluable_operator` naming the op, and `tests/evaluable_core_gate.rs` pins that for all nine gated ops. The note on `CompileError::UnevaluableOperatorError` states why: *"a silent NaN is indistinguishable from a legitimate numerical result and would propagate into the solution."*

The **scalar** interpreter (`Compiled::from_file` / `simulate/`) had no such gate. `resolve_expr` checked only the OPEN tier (`op_registry::check_node`); an evaluable-core op with no rule fell through `eval_op`'s `_ => f64::NAN` backstop and came back as a **number**. Same defect, two author experiences: one names the pipeline stage at fault, the other hands back `actual=NaN expected=25`.

## The change

`resolve_expr` — the one funnel through which every expression the scalar interpreter evaluates becomes a `ResolvedExpr` — now applies a scalar `is_evaluable_op` oracle the same way stage (0) applies the array one:

- **`src/simulate/interpret.rs`** — new `is_evaluable_op`, the scalar analogue of `simulate_array::is_evaluable_op`, kept adjacent to `eval_op` so the two cannot drift. Its `_ => f64::NAN` backstop becomes `unreachable!`, matching the array evaluator's, so a future gap is a loud crate bug rather than a poisoned trajectory.
- **`src/simulate/resolve.rs`** — the gate, right after the existing open-tier check. No `ic` carve-out is needed here (the array path needs one): `flatten` routes `ic(state) = rhs` into `FlattenedSystem::field_ics` as a bare `(target, rhs)` pair and `classify_equations` only ever resolves an equation's RHS, so an `ic` **node** never reaches `resolve_expr`. `an_ic_equation_lhs_still_builds` (which already went through `Compiled::from_file`) still passes.
- **`src/compile_error.rs`** — the message said "in the array interpreter"; both evaluators raise it now, so it no longer names one. Variant shape unchanged.

Because the scalar interpreter's values are `f64`, its rule set is genuinely smaller than the array one's: on top of the nine, it also has no rule for the array/tensor + geometry ops. Those are now refused **by name**. (`is_array_file` routes documents carrying an array op to `simulate_array`, so in practice these are reached by a document built directly for the scalar runner.)

**Correction.** An earlier revision of this description said the same of `const`, `neg` and `true`, and deferred giving the scalar interpreter real rules for them as "a separate divergence worth its own issue". Review showed that is wrong on both counts, so this PR now fixes them instead:

* It is not separable. This PR is what converts that gap from an invisible NaN into a hard refusal, so deferring it means shipping the regression.
* Gating them broke six pre-existing tests and one SHARED cross-binding conformance fixture (see **Fixes made during review** below).
* `evaluate` is a `tier: "stable"` entry in `api-surface.json`, bound in Go, Python and Rust. Go's `esm.Evaluate`, Python's `numpy_interpreter`, Julia and this crate's own ARRAY evaluator all answer for all three. Refusing them would have traded a silent NaN for a diagnostic where four other evaluators return a number.

## Tests

`tests/evaluable_core_gate.rs`, which asserted "no document both VALIDATES and PANICS" for the array runtime, now also asserts **"no document both VALIDATES and silently returns NaN"** for the scalar one, over the same nine ops:

- `no_scalar_document_both_validates_and_returns_nan` — a purely scalar probe document (no `shape`, no array op, no bracketed name, so `is_array_file` routes it to the scalar interpreter) per op. It asserts the document *validates*, that `actual` is not NaN and in fact `None`, and that the message carries `unevaluable_operator` **and the op's name**. `passed == false` alone would not have caught this bug — NaN never equals 25 either, which is how it survived.
  - `enum` carries the one documented exception to the validate half: its operands are an enum name and a symbol (§9.3) and the structural reference checker reads bare string operands as variable refs, so that probe cannot validate. It stays in the list (typed-document route) with the validation half skipped, recorded in the case table rather than dropped.
- `a_nested_unevaluable_op_is_refused_on_the_scalar_path` — the gate reaches nested positions, not just a bare observed body.
- `an_ordinary_scalar_document_still_answers` — the counterweight, so the gate cannot be widened into refusing work the interpreter can do.

Plus unit tests next to the oracle, mirroring the array evaluator's own pair, in `src/simulate/mod.rs`:
`every_registry_op_is_either_evaluable_or_gated`, `the_scalar_evaluable_gap_is_pinned` (the gap pinned member by member), `a_nested_unevaluable_op_is_gated_too`.

## Verified by running

~~Targeted only — CI runs the suites. Nothing here ran `./scripts/test-conformance.sh` or a bare `cargo test`.~~

**That was the defect in this PR's own verification.** A bare `cargo test` on the merged tree found **8 failures** the targeted runs filtered out (652 filtered out of `--lib simulate::tests` alone). See **Fixes made during review**. Current state, on the merged tree:

```
cargo test -j 4 --no-fail-fast    1508 passed; 0 failed; 4 ignored (134 binaries)
cargo clippy --all-targets        no warnings, no errors
cargo fmt --check                 clean
```

Still not run here: `./scripts/test-conformance.sh`, and the Julia / Python / TypeScript / Go suites — `git diff --stat origin/main HEAD` shows the merged branch differs from green `main` in Rust files and `esm-spec.md` only, so those four bindings are byte-identical to `main`.

```
cargo test --test evaluable_core_gate      6 passed; 0 failed
cargo test --lib simulate::tests          18 passed; 0 failed  (652 filtered out)
cargo fmt --check                          clean
```

**Fail-without-fix, verified.** With the `resolve_expr` gate deleted and the `_ => f64::NAN` backstop put back — nothing else changed, the tests untouched — the same run gives:

```
running 6 tests
test an_ordinary_scalar_document_still_answers ... ok
test an_ic_equation_lhs_still_builds ... ok
test a_nested_unevaluable_op_is_refused_on_the_scalar_path ... FAILED
test no_scalar_document_both_validates_and_returns_nan ... FAILED
test every_unevaluable_core_op_ends_in_a_diagnostic_not_a_panic ... ok
test a_true_body_counts_instead_of_panicking ... ok

`skolem` has no scalar evaluation rule and must not evaluate to NaN;
  got actual=Some(NaN) message=actual=NaN expected=25 (rtol=0.000001, atol=0)
a nested `rank` must not NaN the whole expression; got Some(NaN)

test result: FAILED. 4 passed; 2 failed
```

That `actual=NaN expected=25 (rtol=0.000001, atol=0)` is verbatim the symptom the issue reports. The four pre-existing tests still pass without the fix, so the two new ones are detecting this bug specifically and not something incidental. With the fix restored, all six pass and the message is the named `unevaluable_operator` diagnostic.

## Per-binding audit

The issue asked whether the other four bindings' scalar paths have the same hole. **They do not** — Rust was the only one. Every other binding's operator dispatch terminates in a raise/throw; none has a numeric catch-all. No fix was needed elsewhere, so this PR stays Rust-only.

| Binding | Silent-NaN hole? | Terminal arm | Weaker gap worth its own issue |
|---|---|---|---|
| **Rust** | **YES — fixed here** | was `_ => f64::NAN` (`simulate/interpret.rs`) | — |
| **Julia** | No | `throw(TreeWalkError("E_TREEWALK_UNSUPPORTED_OP", op))` in `tree_walk/compile.jl`, `oop.jl`, `access_kernel.jl`; MTK path `throw(ArgumentError)` | `skolem` and `apply_expression_template` slip past the *named* build-time `unlowered_operator` gate (`_op_in_T` returns false; `_cse_hoistable` then routes them into a guardless CSE node) and surface at first RHS evaluation as the generic `E_TREEWALK_UNSUPPORTED_OP`. Loud and deterministic, never a number — a diagnostic-quality gap, not this hole. |
| **Python** | No | `raise NumpyInterpreterError(f"Unsupported op…")` (`numpy_interpreter.py`); `raise SimulationError` (`expression.py`). No NaN sentinel anywhere in the evaluation layer. | The terminal raise is *uncoded* — no `code = "unevaluable_operator"` — and there is no build-time gate at all, only an evaluation-time one. |
| **TypeScript** | No | `throw UnloweredOperatorError` / `throw EvaluatorError('unsupported_operator')` (`codegen.ts`). No `NaN`, no `?? 0`. | Eight of the nine are reported as *open-tier rewrite targets* ("write a rewrite rule"), which is wrong advice for closed-core ops; `{"op":"true"}` is refused although fixtures ship it. |
| **Go** | No | `unevaluableOpError(node)` (`pkg/esm/expression.go`), which already classifies closed-core vs open-tier — the best-behaved binding on this axis. | `distinct` is missing from `closedNonScalarOps` so it gets the open-tier advice; the `enum` arm raises an untyped `fmt.Errorf` and so carries no diagnostic code. Also: AGENTS.md says Go has no runner — true of an ODE *runner*, but Go does ship a public `esm.Evaluate` scalar expression evaluator (pinned `tier: "stable"` in `api-surface.json`), so that table row is inaccurate. |

Follow-ups, none blocking and none included here: the five-way divergence in how the bindings spell this one condition (`unevaluable_operator` / `E_TREEWALK_UNSUPPORTED_OP` / an uncoded `NumpyInterpreterError` / `unsupported_operator` / `unlowered_operator`), plus TypeScript's open-tier misclassification of eight closed-core ops, Go's missing `distinct` and untyped `enum` arm, and the Julia CSE-hoist gap — all recorded as **#281**. The Rust scalar-vs-array `const`/`neg`/`true` asymmetry that an earlier revision listed here is NOT a follow-up: it is fixed in this PR.

## Interaction with PR #209

PR #209 (`claude/issue-188-table-lookup-lowering`, the `table_lookup` lowering this issue came out of) is **still open** and not in `origin/main`. The two touch two files in common:

- `src/compile_error.rs` — #209 adds `table_lookup` diagnostics; this PR rewords the `UnevaluableOperatorError` message. Different lines.
- `tests/evaluable_core_gate.rs` — #209 edits lines 15-24 and 71-77 of the module doc comment (moving `table_lookup` from "lowered at LOAD" to "lowered on the way into the BUILD"). This PR appends a new `# The scalar half` section after line 32 and new tests at the end of the file, deliberately leaving those lines alone.

Expected to merge cleanly in either order. Once #209 lands, `table_lookup` reaches the scalar gate here for the same reason it reaches the array one — the probe declares no `function_tables`, so there is nothing to lower it against — which is what the new test already exercises.

## Base

Branched from `origin/main` @ 71e25b380. `main` has since advanced to 9cb68fb40; of the files this PR touches, only `src/simulate_array/compile.rs` also changed there, in unrelated regions (lines ~19/367/394/3344 vs. the comment at ~2704), so the merge is expected to be clean and CI will build the merged state.

---

# Fixes made during review

`main` had moved 221 commits. The merge (`b2aa703bc`) hit one textual conflict — main renamed `run_pde_tests` → `run_inline_tests`, which all three new tests call — and then a bare `cargo test` on the merged tree gave **1506 passed, 8 failed**:

* **`neg` refused** (5 tests): `problem::stepping_tolerance_tests::*` ×4 and `solver_block::a_typed_file_carries_its_block_onto_the_problem` all build `D(y) ~ {"op":"neg","args":["y"]}` — the canonical unary negation `canonicalize.rs` emits and never rewrites away.
* **scalar `const` refused** (1 test): `function_tables_inline_tests::a_table_lookup_observed_evaluates_like_its_hand_lowered_twin`, driving the shared fixture `tests/conformance/function_tables/inline_test/fixture.esm`, whose first equation is `D(x) ~ {"op":"const","args":[],"value":0.0}` — the very document issue #220 was reported from. That RHS was **silently NaN before this PR**, unnoticed because no assertion reads `x`: a live instance of the #220 defect class, now actually fixed.
* **`unreachable!` panicking on public input** (2 tests): `tests/interpret.rs::{unknown_op_returns_nan, differential_ops_on_rhs}` hand-build a `ResolvedExpr::Op`, disproving the claim that `resolve_expr` is its only constructor.

`53fded9a2` — `neg` → `-v(0)` and `true` → `1.0` arms in `eval_op`; a scalar `const` folds to `ResolvedExpr::Number` in `resolve_expr` (its value lives on the node, not in `args`, so a plain `Op` cannot carry it — an ARRAY `const` stays gated by name). New test `neg_true_and_a_scalar_const_evaluate_as_the_other_bindings_do` fails if any of the three is reverted.

`f9b58dde0` — the repo owner's three rulings on the review:

1. **`unreachable!` restored, and made sound.** `ResolvedExpr::Op` is sealed `#[non_exhaustive]` and built out-of-crate only through the new checked `ResolvedExpr::op(name, args) -> Result<_, CompileError>`. **BREAKING** at `extension` tier; the owner accepts it. `api-surface.json` needs no regeneration (it pins crate-root `pub use` names, not variant shape). `unknown_op_returns_nan` is deleted; `differential_ops_on_rhs` keeps its `D`-is-0.0 half and its spatial-sugar half becomes `an_op_with_no_rule_cannot_be_built`, asserting the stronger property the seal buys.
2. **Gate ordering kept, and pinned.** The evaluability gate stays ahead of the §11.3 Float32 gate — the scalar path cannot evaluate the geometry ops in any precision, so `float32_unsupported` would send the author to fix the wrong thing. `the_evaluability_gate_precedes_the_float32_gate` pins it under Float32, where it is observable.
3. **`unevaluable_operator` is now normative.** esm-spec §9.6.6 gains a row next to `unlowered_operator` stating how the two differ: outside evaluable-core and never lowered, versus inside evaluable-core with no rule in *this* evaluator.


ctessum:
**Merge-order note: this PR and #251 disagree about `neg`.**

Raised by the agent working #206 (PR #251), which adds the missing `neg` arm to the scalar interpreter — the same silent-NaN hole this PR closes, found from the other direction.

This PR's `is_evaluable_op` explicitly lists `neg` as **not** evaluable, with the comment:

> `const`, `neg` and `true`, which the array runtime evaluates and this one never had an arm for.

That is accurate as of this PR. But `neg` is registered evaluable-core, so gating it makes the two Rust runners disagree about a core op — and #251 gives the scalar interpreter the arm it was missing.

**Whichever of the two merges second needs one edit:** add `| "neg"` to the `matches!` list and drop `neg` from that comment. Nothing else conflicts between the two PRs.

Separately, and worth a look while this is open: the same list keeps `"D"` evaluable on "legacy-parity zero" grounds. #251 makes that stale — after it, an RHS `D` cannot reach a compiled model, and `fold_constant_expr` is the last direct-evaluation entry that would still fold it to `0`. Under #251's amended esm-spec §4.2, an unresolvable `D` must be reported rather than given a value, "in particular not `0`".

PR #283, opened 2026-09-09
perf(oop): the sub-kernel descriptors were the ue readers, not the ghost gathers

ctessum-claude:
## What this is

The `ess-oop-ssa` spike (`src/tree_walk/SSA_SPIKE.md`) redirects a consumer
access-kernel descriptor's read from the flat extended buffer `ue` to its
PRODUCER'S VALUE, and skips a producer's scatter into `ue` when static
accounting shows nothing reads its block off the buffer any more. It shipped
with four of its own "What blocks the rest" items open. This closes three of
them — and, more usefully, **measures which one mattered. The list's order was
wrong.**

Independent of #273 (`perf/contraction-tier-order`) and #278
(`perf/cross-shape-state-gather`); based on `origin/main`, touching only the
`_OopSSA*` sections of `oop.jl`, that file's test, and the design note.

## The measurement that redirected the work

`oop_ssa_stats` now carries a per-reason **blocker tally**: for each producer
whose scatter survived, which read surface still reads its block. On the ReSEACT
transport RHS at 6x6x8 — the target of the whole exercise — it says:

* **ghost-masked `_AK_STATE_TBL_BOX` gathers, blocker #1 and the note's primary
  target, do not occur in this build.** `n_ghost_edges = 0` on both split parts,
  and none in any 1-D host fixture either. The arm is implemented and
  unit-tested anyway (it is cheap and it is correct), but it buys nothing here.
* **the template SUB-KERNEL descriptors (blocker #2) are the read surface**:
  885 of the 985 candidate descriptors and **1 074 487 of the 1 106 635
  candidate read elements — 33x the top-level edges' volume** — and they held
  9 of the 11 surviving producer scatters alive.
* E-lane CSR gathers (#3) and `_AK_STATE_FIXED` pins are 0 here too.

## What changed

Three redirect arms, each with a bisect knob (`ESS_OOP_SSA_SUB`,
`ESS_OOP_SSA_PGATHER`, `ESS_OOP_SSA_GHOST`, all default ON inside
`ESS_OOP_SSA=1`, so one process can A/B an arm and every engagement test has a
negative control):

1. **Per-sub descriptor tables** (`_OopSSAKernel.subs`, `_oop_ssa_subctx`).
   A sub-kernel is evaluated at the PARENT's lanes and
   `_build_oop_desc_vectors` resolves its descriptors against that same
   enumeration, so its gathers decompose by exactly the same
   slot→(producer, position) map; only the table differs (`S.acc`, not
   `K.acc`), which is why the spike's per-kernel table could not carry them.
2. **Tier 2b, the single-producer VALUE gather.** Past the slice-worthwhileness
   bound a mapping that lies inside ONE producer now gathers that producer's
   value at producer-local positions: the same single op the dense `ue` gather
   was, over an index vector of the same length. The win is not op count — it is
   that the read stops touching the flat buffer, which is what lets the scatter
   go.
3. **Ghost-masked table gathers.** A masked lane's value is discarded by the
   emitter's `ifelse.(mask, 0, ·)` select, so it is a WILDCARD for the
   decomposition: fill it with whatever continues a neighbouring run (forward,
   then backward, then a placeholder). A one-sided boundary stencil collapses to
   a single slice, and the select is the op it already was.

`_ssa_decompose` and its masked variant are unified into `_ssa_lane_owners`
(resolve lanes to owners) + `_ssa_pack_ref` (pick the cheapest form), so all
three surfaces and all four tiers go through one path.

## Coverage

Host fixtures (`oop_ssa_stats`, the note's four): unchanged at 100% — they were
already all tier 1/2, which is the point of the `none` row below.

ReSEACT, per arm, `tools/diag/p12_ssa_blockers.jl` at 6x6x8:

**transport (split part 1)**

| arms | edges | sub-kernel | scatters skipped | blocked by |
|---|---|---|---|---|
| none (the spike) | 51/100 | 0/885 | 6/17 (720 slots) | sub 9, frag 5, scalar 2, scan 1 |
| ghost only | 51/100 | 0/885 | 6/17 | unchanged |
| sub only | 51/100 | 5/885 | 7/17 | sub 8, frag 5, scalar 2, scan 1 |
| tier 2b only | **97/100** | 0/885 | 7/17 | sub 9, scalar 2, scan 1 |
| **all three** | **97/100** | **731/885** | **13/17 (4 874 of 8 042 slots)** | sub 2, scalar 2, scan 1 |

**chemistry (split part 2)**

| arms | edges | scatters skipped |
|---|---|---|
| none | 561/563 | 57/59 |
| **all three** | **563/563** | **59/59 — the chemistry RHS no longer scatters into `ue` at all** |

The two arms that pay are **synergistic**: the sub tables make the descriptors
visible, tier 2b gives them a form. 731 of the 885 need both, because a
sub-kernel gather is almost always too fragmented for slices (sub alone
redirects 5).

What is left on transport is blocker #4 (scalar-walker reads: 2 producers, and
`blockers_only` says BOTH are blocked by nothing else), one sub-kernel
descriptor group, and one level scan fold.

## Exactness

Unchanged, and it is the contract. A redirect returns exactly the array the
scatter would have copied — or a slice, or a gather, of it — and for a
ghost-bearing descriptor the select overwrites precisely the lanes whose
positions were invented, so no invented value can reach arithmetic (nor a
derivative: `ifelse` propagates only the selected branch).

* `:oop`(on) == `:oop`(off) == `:inplace` at Float64 `==`, and ForwardDiff
  jacobians `==`, on every fixture in `tree_walk_oop_ssa_test.jl`.
* With the flag OFF every structure is the empty singleton and the emitted
  program is byte-for-byte the pre-spike emitter.
* With all three knobs at 0 the coverage is bit-identical to the spike's — the
  `none` row above is the assertion.

## Tests

121 new assertions in `test/tree_walk_oop_ssa_test.jl`, all with negative
controls:

* white-box `_ssa_lane_owners`: the level and ownership refusals, out-of-range
  slots, the ghost wildcard fill (middle ghost → ONE slice; leading ghost that
  cannot back-extend → placeholder + slice; leading ghost that can → one slice),
  that every invented position stays inside its producer's value for all 14
  non-trivial 4-lane masks, and the all-ghost decline;
* white-box `_ssa_pack_ref`: the tier choice, that a fragmented single-producer
  mapping takes tier 2b at exactly the producer-local positions, that
  `ESS_OOP_SSA_PGATHER=0` sends it back to the buffer, and that a fragmented
  MULTI-producer mapping has no one-op form either way;
* end-to-end **tier 2b**: a reversed observed read (`g[N+1-i]`, L runs of
  length 1). Bit-identical to flag-off, to `ESS_OOP_SSA_PGATHER=0`, and to
  `:inplace`; ForwardDiff `==`; engagement (`n_fast` and
  `n_skipped_scatters` both strictly greater than the arm-off build, and the
  producer's scatter is the one that dies);
* end-to-end **sub-kernel tables**: the duo-rule shape from
  `codegen_subcall_fn_test.jl` (the shape that mints `_NK_SUBCALL`s at all) —
  26 sub descriptors, all redirected, `0` with `ESS_OOP_SSA_SUB=0`, values
  bit-identical across all three builds and `:inplace`.

Suite with the flag **OFF** (the default path): `tree_walk_oop` 164,
`tree_walk_oop_ssa` 45 + 121 new, `oop_merge` 20 + 55, `array_obs_materialize`
71 + 3, `observed_materialization` 4, `scan_prefix` 333, `oop_scalar_batch` 42,
`tree_walk_observed_slots` 8 + 12 + 36, `tree_walk_iip_generic` 69,
`codegen_subcall_fn` 50 — all green.

## Not verified here

The ReSEACT timing A/B and the whole-extended-buffer copy census need the
offline forcing environment; they live in the consuming repo
(`tools/diag/p13_ssa_ab.jl` — which includes the driver twice, once per arm,
because these are BUILD-time flags and cannot ride a trace-time `Ref` the way
`ess-oop-levelbase` did — and `tools/diag/p11_ue_traffic.py`).

---

# ReSEACT at CONUS, measured (2026-09-09)

4x5 CONUS = 13x7x72 = 6 552 cells. Two driver builds in ONE process, arms
interleaved A/B/A/B on the same inputs with fresh device buffers per call, and
every pair run in **both arm orders** — which reproduced to three digits, so
"which arm compiled first" is not what is being measured.
`tools/diag/p13_ssa_ab.jl` + `tools/diag/p11_ue_traffic.py`.

## Timing

| program | arms off | arms on | off/on | exactness |
|---|---|---|---|---|
| `ros_step` (chemistry step) | 38.90 / 38.62 ms | 28.23 / 28.19 ms | **1.38 / 1.37 faster** | **bit-for-bit** |
| `ros_vjp` (chemistry VJP) | 81.11 / 81.50 ms | 56.71 / 56.83 ms | **1.43 / 1.43 faster** | λ **bit-for-bit**; p-grad max abs 2.5e-24 on a 7e-5 scale (3.5e-20 scale-relative), 0 of 162 components over 1e-9 |
| `rhs` (transport RHS) | 3.755 / 3.771 ms | 6.405 / 6.372 ms | 0.59 / 0.59 **slower** | see below |
| `rhs_vjp` | 80.5 / 80.0 ms | 72.2 / 72.4 ms | **1.12 / 1.11 faster** | |
| `ssp_step` (4-stage transport) | 13.41 / 13.20 ms | 24.25 / 24.13 ms | 0.55 / 0.55 **slower** | 5.55e-15 pointwise relative |
| `ssp_vjp` | 306.6 / 307.4 ms | 254.9 / 255.6 ms | **1.20 / 1.20 faster** | |

Weighted by the adjoint's own step mix at CONUS (45.3 chemistry + 3.2 transport
steps per 300 s window), the per-window cost goes **6.46 s → 4.74 s, 1.36x** —
and the chemistry half, which is 84% of it, is exact.

## Whole-buffer copies and write traffic

| program | copies off → on | MB | real element writes off → on |
|---|---|---|---|
| `rhs` | 2 → 3 | 4.0 → 5.9 | 2.49 M → 2.69 M |
| `rhs_vjp` | **98 → 82** | 194.0 → 162.3 | 59.55 M → 55.22 M |
| `ssp_step` | 8 → 12 | 15.8 → 23.8 | 10.43 M → 16.97 M |
| `ssp_vjp` | **394 → 331** | 779.9 → 655.2 | 231.8 M (1854 MB) → 176.3 M (1411 MB) |

The `off` column reproduces the campaign's published baseline on the nose
(`rhs_vjp` 98 vs the recorded 99, `ssp_vjp` 394 exactly), so the two arms really
are the spike and the spike-plus-extension. `ssp_vjp` loses 63 of its 394
whole-buffer copies and **24% of its write traffic**, and runs 1.20x — against
the ~1.8x ceiling if every copy vanished.

## Why the transport forward loses while everything else wins

A producer's `dynamic_update_slice` into `ue` **aliases its operand in the
forward**: XLA:CPU writes only the update, so the scatter is nearly free there,
and skipping it instead forces the producer's value to be materialized as its
own buffer. On the PPM transport RHS — 17 producers, output blocks up to 3 456
elements — that trade is a loss (`ssp_step` +63% element writes, 1.8x slower).
On the chemistry RHS it is a large win in both directions, because there the
redirect turns fragmented gathers of the 12 326-element buffer into gathers of
much smaller producer values and 59 of 59 scatters disappear.

The copies the scatter chain causes are all in the **reverse**, where each `ue`
version has ~70 live consumers and copy-insertion duplicates the whole buffer
per version. That is where removing readers pays on both halves.

This is also exactly why `ess-oop-levelbase` failed where this does not: it
reduced the number of `ue` VERSIONS — which the forward already got for free —
instead of removing READERS, which is what frees the reverse. Its census went
99 → 102 copies; this one goes 98 → 82 and 394 → 331.

## Exactness at CONUS

Host is exact (`==`) and asserted in the suite. Traced, chemistry is exact and
transport differs only by XLA re-fusing a changed operand graph (FMA and
vectorization order):

| program | max abs difference | ÷ scale | components over 1e-9 of scale |
|---|---|---|---|
| `ros_step` | **0** | 0 | 0 of 85 176 |
| `ros_vjp` λ | **0** | 0 | 0 of 85 176 |
| `ros_vjp` p-grad | 2.5e-24 | 3.5e-20 | 0 of 162 |
| `rhs` | 1.9e-17 | 1.6e-14 | 0 of 85 176 |
| `rhs_vjp` λ | 1.7e-21 | 1.5e-16 | 0 of 85 176 |
| `ssp_step` | 5.6e-15 | **1.1e-18** | 0 of 85 176 |
| `ssp_vjp` λ | 3.5e-18 | 3.2e-16 | 0 of 85 176 |

Not one of the 85 176 state or λ components in any program differs by more than
1e-9 of that quantity's own scale. The POINTWISE relative metric is what
initially read `2.0` on the VJPs and `5.55e-15` on `ssp_step`: it is
uninformative on components at ~1e-310 (rel 2.0 is the signature of an
opposite-signed denormal), which is why `P13_SCAL` reports absolute and
scale-relative figures. The only entries with any component over the 1e-9
threshold are the p-gradients of the SINGLE-STEP transport programs, whose whole
gradient vector is ~1e-19 in magnitude because θ barely enters one transport
step — the absolute differences there are 5e-29 and 8e-20.

## Suite

Full `Pkg.test()` with `ESM_TEST_REACTANT=1`, both arms:

* flag **OFF**: **89 034 pass, 8 broken, 89 042 total**, 70 min, exit 0 — the
  default path is untouched.
* forced **`ESS_OOP_SSA=1`**: **89 029 pass, 5 fail, 8 broken, 89 042 total**,
  64 min. All 5 failures are `reactant_oop_intern_test.jl` lines 155, 161, 167,
  168 and 191 — precisely the 5 interning-ENGAGEMENT assertions (`hits == 0`;
  `ops["1"] < ops["0"]` reading 210 < 210) that `SSA_SPIKE.md` already documents
  as the two features composing, with every VALUE assertion passing. Nothing
  else in 89 042 tests fails.

`reactant_oop_test.jl` 35/35 and `reactant_oop_intern_test.jl` 19/19 also pass
standalone with the flag off.

## Recommendation

On this evidence `ESS_OOP_SSA=1` (with these arms, which ride inside it) is
worth defaulting: the dominant loop is 1.36x cheaper and the chemistry half is
exact. Two caveats a maintainer should weigh:

1. flag-on is not bit-identical to flag-off in the traced TRANSPORT path
   (5.6e-15), so defaulting it changes numbers for existing traced users and
   should be a noted change, not a silent one;
2. the transport forward is 1.8x slower, so a caller that runs the transport
   primal without a reverse is worse off. If that matters, the arms already
   have per-arm knobs and the emitter could gate the scatter-skip on producer
   block size — but on the adjoint workload this PR targets, the mix is
   decisively favourable.

### DIFF
--- a/pkg/EarthSciAST.jl/src/tree_walk/SSA_SPIKE.md
+++ b/pkg/EarthSciAST.jl/src/tree_walk/SSA_SPIKE.md
@@ -180,14 +180,58 @@ fold.
   forcing reads go through the buffers argument, never `ue` — and the
   discrete-cadence refresh stays visible with fill scatters skipped (probed).
 
+## ReSEACT at CONUS: 1.36x on the whole adjoint loop, chemistry exact
+
+Measured 2026-09-09 at 4x5 CONUS (13x7x72, 6 552 cells), two driver builds in
+ONE process, arms interleaved, every pair run in BOTH arm orders — reproducing
+to three digits, so the order is not what is being measured.
+`tools/diag/p13_ssa_ab.jl` + `tools/diag/p11_ue_traffic.py` in the consuming
+repo.
+
+| program | arms off | arms on | off/on | copies | real element writes |
+|---|---|---|---|---|---|
+| `ros_step` (chemistry) | 38.90 ms | 28.23 ms | **1.38** (bit-for-bit) | | |
+| `ros_vjp` | 81.11 ms | 56.71 ms | **1.43** (λ bit-for-bit) | | |
+| `rhs` (transport RHS) | 3.755 ms | 6.405 ms | 0.59 | 2 → 3 | 2.49 M → 2.69 M |
+| `rhs_vjp` | 80.5 ms | 72.2 ms | **1.12** | **98 → 82** | 59.55 M → 55.22 M |
+| `ssp_step` (4-stage) | 13.41 ms | 24.25 ms | 0.55 | 8 → 12 | 10.43 M → 16.97 M |
+| `ssp_vjp` | 306.6 ms | 254.9 ms | **1.20** | **394 → 331** | 231.8 M → 176.3 M |
+
+Weighted by the adjoint's step mix (45.3 chemistry + 3.2 transport steps per
+300 s window) the per-window cost is 6.46 s → 4.74 s, **1.36x**.
+
+WHY THE TRANSPORT FORWARD LOSES. A producer's `dynamic_update_slice` into `ue`
+ALIASES its operand in the FORWARD, so XLA:CPU writes only the update and the
+scatter is nearly free there; skipping it instead forces the producer's value to
+be materialized as its own buffer. On the PPM transport RHS (17 producers,
+blocks up to 3 456 elements) that trade is a loss. On chemistry it is a win in
+both directions, because there the redirect replaces fragmented gathers of the
+12 326-element buffer with gathers of much smaller producer values and 59 of 59
+scatters disappear.
+
+The copies the scatter chain causes are all in the REVERSE, where each `ue`
+version has ~70 live consumers and copy-insertion duplicates the whole buffer
+per version. That is where removing readers pays on both halves: `ssp_vjp`
+loses 63 of its 394 whole-buffer copies and 24% of its write traffic for 1.20x,
+against a ~1.8x ceiling if every copy vanished.
+
+It is also why `ess-oop-levelbase` (one read version per level) failed where
+this succeeds: it reduced VERSIONS, which the forward already got for free,
+instead of removing READERS. Its census went 99 → 102 copies; this one goes
+98 → 82 and 394 → 331.
+
 ## What is NOT verified
 
-The `ESM_TEST_REACTANT=1` traced census (`reactant_oop_ssa_test.jl`) asserts the
-DUS delta equals `n_skipped_scatters` on the toy fixtures; the extension's arms
-are exercised there only insofar as those fixtures engage them. The ReSEACT
-timing A/B and the whole-buffer copy census live in the consuming repo
-(`tools/diag/p13_ssa_ab.jl`, `tools/diag/p11_ue_traffic.py`) because they need
-the offline forcing environment.
+Bit-identity across the arms is asserted and holds on HOST (`==`) and, on the
+toy fixtures, in the traced census (`reactant_oop_ssa_test.jl`, which pins the
+DUS delta to `n_skipped_scatters`). At CONUS the two arms are NOT bit-identical:
+changing the operand graph changes XLA's fusion, hence FMA and vectorization
+order. chemistry `ros_step` is bit-for-bit and `ros_vjp`'s λ is bit-for-bit (its
+p-gradient differs in the 20th significant figure); transport `ssp_step`
+differs by 5.6e-15 pointwise relative. The pointwise metric on the VJPs is
+uninformative — it reads exactly 2.0 on components at ~1e-310 — so the probe
+also reports a scale-relative figure and the count of components over 1e-9 of
+scale.
 
 ## Generalization assessment (honest)
 
--- a/pkg/EarthSciAST.jl/src/tree_walk/SSA_SPIKE.md
+++ b/pkg/EarthSciAST.jl/src/tree_walk/SSA_SPIKE.md
@@ -25,8 +25,24 @@ With the flag on, a consumer descriptor references its producer's RESULT VALUE:
 * **tier 2 (slice)** — the gather decomposes into few consecutive-position runs,
   each inside one producer ⇒ one `slice` per run (+ one `concatenate` when
   more than one run);
+* **tier 2b (producer gather)** — the mapping lies inside ONE producer but
+  shatters into more runs than slices are worth ⇒ one `gather` of that
+  producer's value at producer-local positions. Exactly the single op the dense
+  `ue` gather was, over an index vector of the same length: the win is not op
+  count, it is that the read no longer touches the flat buffer;
 * **tier 3 (fallback)** — everything else keeps the dense gather from `ue`,
   byte-for-byte.
+
+Three descriptor surfaces feed those tiers: a kernel's own top-level
+descriptors, its TEMPLATE SUB-KERNEL (`_NK_SUBCALL`) descriptors — resolved by
+`_build_oop_desc_vectors` against the same parent lanes, so decomposable by the
+same map, but indexed by `S.acc` and therefore carried in a per-sub table — and
+GHOST-MASKED `_AK_STATE_TBL_BOX` descriptors, where a masked lane's value is
+discarded by the emitter's `ifelse` select and is therefore a WILDCARD the
+decomposition fills with whatever continues a neighbouring run. Each of the
+three has a bisect knob (`ESS_OOP_SSA_SUB`, `ESS_OOP_SSA_PGATHER`,
+`ESS_OOP_SSA_GHOST`, all default ON inside `ESS_OOP_SSA=1`), which is what lets
+one process A/B an arm and what each engagement test's negative control flips.
 
 A producer's scatter into `ue` is emitted only when static accounting finds a
 residual reader of its block; otherwise it is skipped and the value flows only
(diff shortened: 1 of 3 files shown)


ctessum-claude:
Reviewed as a group with #292 and #294 against current `main`, not the branch
point. **Merging this, with #294 immediately after it. #292 is closed.**

## Merge safety, checked rather than assumed

`main` moved 162 commits since `3aa046d65`. Exactly one of them touches
`oop.jl` (`70a7d57e1`), and it wraps `_build_oop_acc_plan` *additively* rather
than renaming it away — so the class of semantic merge conflict that bit most of
the last batch is absent here. I also diffed the two structures this plan
consumes, `_OopAccPlan` and `_AccDesc`, between the fork point and `main`: both
byte-identical. Nothing the SSA plan reads has moved underneath it.

All three merged textually clean into `main`. The feature is entirely behind
`_oop_ssa_enabled()`, which is `get(ENV, "ESS_OOP_SSA", "") == "1"` — default
off, with `_build_oop_ssa_plan` returning the empty singleton before doing any
work — so the default path is untouched and CI exercises it as such.

I read the ghost wildcard fill closely, since inventing positions is the one
place this could quietly corrupt arithmetic. The bounds hold on all three
branches: forward extension guards `pos + 1 <= prod_len[pl]`, backward guards
`pos - 1 >= 1`, and the placeholder falls back to pid 1 position 1 behind a
`prod_len[1] >= 1` check. Every invented position stays inside its producer's
value, which is what the select then discards anyway.

## One correction pushed (`91e81c132`)

The section heading claimed **"1.36x on the whole adjoint loop"**. It is not a
loop figure: the paragraph under it derives 1.36x from a per-window step+VJP
mix, which prices one primal against one VJP — while the adjoint evaluates the
primal twice per accepted step (forward step, backward replay), and the loop
also carries the refresh and host terms. #294's own table prices the same arm at
**1.19x** on the loop, 130 lines below a heading that still said 1.36x.

That gap matters in the unflattering direction: the transport primal is the half
that regresses, so the mix understates exactly the cost this arm incurs. The
heading now names the quantity and says outright that it is not a loop figure.

## Not verified here

The ReSEACT timings need the offline forcing environment and I took them on the
PR's word. They also predate #278, which moves the descriptor population this
plan resolves — filed as #296. That is a calibration question, not a soundness
one: I confirmed both descriptor kinds are already enumerated by the residual
marking, so no read surface goes unmarked.

Issue #272, opened 2026-09-08
cadence: the conformance oracle and Julia's mirror seed observeds from the strict bare-LHS set, so a state-free ARRAYED observed never folds at bind (§6.3.1, §5.7.2)


ctessum:
## Summary

The cadence pass's leaf-seed table is stated in terms of esm-spec §6.3.1's
classification functions precisely so five bindings cannot disagree about which
nodes fold. Two implementations of it — the **conformance reference oracle** and
**Julia's mirror of it** — do not call those functions. They re-derive "is this
an observed?" locally from `isinstance(lhs, str)` / `isa(lhs, AbstractString)`:
the strict bare-LHS **inlining** set, used as if it were the classification,
which §6.3.1 says it is not ("it does not narrow the partition").

This is the same substitution as #232 (Julia tree-walk), #231 / #237 (Python
flatten) and #268 (Go and TypeScript classification), in a fourth place.

## The two sites

`scripts/run-cadence-conformance.py:127-146` — the reference oracle:

```python
def _observed_definitions(model: dict) -> dict:
    """...An observed unknown is one an equation defines with a BARE-VARIABLE LHS
    (esm-spec §6.3.1)...."""
    ...
    for eq in model.get("equations", []) or []:
        lhs = eq.get("lhs")
        if isinstance(lhs, str) and lhs in unknowns and lhs not in defs:
            defs[lhs] = eq.get("rhs")
```

`pkg/EarthSciAST.jl/src/cadence.jl:151-166` — Julia's mirror, same gate:

```julia
        lhs = get(eq, "lhs", nothing)
        isa(lhs, AbstractString) || continue
```

Both docstrings say "a BARE-VARIABLE LHS". §6.3.1 says the criterion is the LHS's
**base name**, bare *or indexed*, at any rank, and it retired the "bare-variable
LHS" wording explicitly:

> Earlier drafts of this table wrote the observed criterion as "a bare-variable
> LHS", which was written for scalar equations and contradicts the semantic
> criterion in the arrayed case … The semantic criterion governs.

Julia's own `seed_leaf` docstring states the rule these two then break:

> every binding MUST seed from those functions rather than re-deriving the
> categories locally — five local derivations are five chances to disagree about
> which nodes fold

## Measured divergence

On the fixture added by PR #268,
`tests/conformance/classification_indexed_lhs/fixtures/observed_indexed_lhs.esm`,
[...]

Issue #270, opened 2026-09-08
flatten: an arrayed observed belongs in BOTH state_variables and observed_variables (esm-libraries-spec §4.7.5); no binding does that, and they miss in two opposite directions


ctessum:
## The contract

`esm-libraries-spec.md` §4.7.5's flattened-representation table states, for the
two buckets, that an **arrayed observed is in BOTH**:

> | `state_variables` | … Differential unknowns …, PLUS `algebraic_variables`, PLUS **any arrayed observed that materializes into a buffer**. |
> | `observed_variables` | Unknowns DEFINED by an equation, **bare-LHS or indexed-LHS** (esm-spec §6.3.1). A scalar observed is eliminated by substitution and is NOT in `state_variables`; **an arrayed observed materializes into a buffer and IS**. |

So for an arrayed observed `wf` defined by `wf[k] ~ 2*k`, the spec asks for
`wf ∈ state_variables` (the solver must allocate the buffer) **and**
`wf ∈ observed_variables` (an equation defines it).

## No binding does that, and they fail in two opposite directions

Measured on PR #250's committed fixture
`tests/conformance/pde_inline_observed_indexed_lhs/fixtures/observed_indexed_lhs.esm`,
whose `wf` and `ws` are array unknowns defined by the indexed LHS
`aggregate{k}(index(wf, k)) ~ …`:

| Binding | `state_variables` | `observed_variables` | vs. §4.7.5 |
|---|---|---|---|
| Julia (`main`) | `[M.u, M.v]` | `[M.wf, M.ws]` | missing from `state_variables` |
| Python (`main`) | `[M.u, M.v, M.wf, M.ws]` | `[]` | missing from `observed_variables` |
| Python (@ PR #237) | `[M.u, M.v]` | `[M.wf, M.ws]` | missing from `state_variables` |
| Rust (`main`) | `[M.u, M.v, M.wf, M.ws]` | `[]` | missing from `observed_variables` |
| Go (@ PR #268) | `[M.u, M.v, M.wf, M.ws]` | `[]` | missing from `observed_variables` |
| TypeScript (@ PR #268) | `[M.u, M.v, M.wf, M.ws]` | `[]` | missing from `observed_variables` |

All six rows are measured, not read off the source. Python `main` is the
pre-#237 baseline; the `@ PR #237` row is that branch's `pkg/earthsci-ast-py/src`
on `PYTHONPATH`. Go and TypeScript are measured with PR #268 applied — that PR
[...]

PR #290, opened 2026-09-09
flatten: an arrayed observed is in BOTH state_variables and observed_variables (§4.7.5)

ctessum-claude:
Fixes the defect in #270: `esm-libraries-spec.md` §4.7.5 step 4 puts an arrayed
observed in **both** `state_variables` and `observed_variables`, and no binding
did that.

## The §4.7.5 reading

The step-4 field table's two unknown maps are **not** a partition:

> | `state_variables` | … Differential unknowns …, PLUS `algebraic_variables`, PLUS **any arrayed observed that materializes into a buffer**. |
> | `observed_variables` | Unknowns DEFINED by an equation, **bare-LHS or indexed-LHS** (esm-spec §6.3.1). A scalar observed is eliminated by substitution and is NOT in `state_variables`; **an arrayed observed materializes into a buffer and IS**. |

`state_variables` is the SOLVED-FOR VECTOR; `observed_variables` is the §6.3.1
CLASSIFICATION. §4.7.5 says so in as many words a few lines above the table
("**`state_variables` is the solved-for vector, not a classification**"). Dual
membership is not a new mechanism: the same table already requires it of an
algebraic unknown ("A **subset** of `state_variables` — a DAE solves for them"),
and `im` already appears in both `state_variables` and `algebraic_variables` in
every binding.

**Which observeds materialize.** §4.7.5's qualifier — "any arrayed observed *that
materializes into a buffer*" — is read through §6.3.1's own distinction between
the classification and the narrower set it sanctions for **inlining
specifically** (which "does not narrow the partition"): an observed defined by a
**bare-variable LHS** is eliminated by substituting its definition into every
consumer, so it occupies no buffer, while one defined by an **indexed LHS**
(`y[i] ~ …`, or the `aggregate{k}(y[k]) ~ …` shell) is arrayed by construction
and its consumers index the buffer instead. So the materialized set is
`observed_unknowns \ inlined_unknowns` — which is the set Rust's, Go's and
TypeScript's own comments already named "materializes into a buffer".

## Before / after

Measured on `tests/conformance/classification_indexed_lhs/` (#268's fixture),
whose `wf`/`ws` use the `aggregate{k}(index(w, k))` spelling, `wb` the bare
`index(wb, i)` of §6.3.1's worked example, `sc` is the scalar control, `u`/`v`
are ODE states under the same `aggregate` shell and `im` is algebraic:

| Binding | `state_variables` | `observed_variables` |
|---|---|---|
| **before** — Python | `[u, v, wb, im]` | `[wf, ws, sc]` |
| **before** — Julia | `[u, v, im]` | `[wf, ws, wb, sc]` |
| **before** — Go / TS / Rust | `[u, v, wf, ws, wb, im]` | `[sc]` |
| **after** — all five | `[u, v, wf, ws, wb, im]` | `[wf, ws, wb, sc]` |

(`algebraic_variables` is `[im]` throughout.)

### Two findings the issue did not have

**Julia is a THIRD corner, not one of the two #270 measured.** #270 measured two
directions — Python dropping arrayed observeds from `state_variables`, and
Go/TS/Rust dropping them from `observed_variables`. Julia is neither: it gated on
the broad `observed_unknowns`, so an arrayed observed fell out of the solved-for
vector *entirely* (`state_variables = [u, v, im]`). That is the corner Rust's
own source comment warns about by name — "the same class of defect
esm-libraries-spec 45fa534a0 corrected for `algebraic_variables`" — and it is the
most consequential of the three, because a solver built from that flattened form
never allocates the buffer at all.

**Julia's flatten conformance driver was actively skipping the comparison, with
a comment admitting nobody implemented §4.7.5.** `_FC_DIVERGENCES` listed
`state_variables` and `observed_variables` on `edge_enumeration_area_eff`, and
the header entry justifying the skip read:

> The §4.7.5 step-4 table says BOTH — an arrayed observed is observed AND
> materializes into the solved-for vector — so this is the one case that could
> pin that rule and the corpus cannot, because the oracle does not implement the
> observed half. **Neither binding implements the table's full reading today.**

So the one shared fixture that *could* have caught this had its comparison
switched off, in the binding that was furthest from the spec, with the reason
written down and then left. Removing that skip is arguably the most valuable
line in this PR: it turns `edge_enumeration_area_eff` back into a live
cross-binding pin for exactly the rule #270 is about, and it is why the revert
check below fails in Julia at all.

## What changed, per binding

| Binding | Site | Was |
|---|---|---|
| Julia | `namespacing.jl` `_collect_model!` | gated on the broad `observed_unknowns`; adds `_inlineable_observed_unknowns` to split off the strict form |
| Python | `flatten.py` `_collect_model` | read `inlined_unknowns` as if it were the classification; now reads both, from the model **as written** |
| Rust | `flatten.rs` `build_model_block` | `if class.is_inlined(…)` alone |
| Go | `flatten.go` `collectModel` | `inlinedUnknownSet` alone |
| TypeScript | `flatten.ts` `collectModel` | `observedDefinitions(…, {bareOnly: true})` alone |

A binding files **one** `FlattenedVariable` record in both maps and keeps the
§6.3.1 role tag (`observed`) in each — the role follows the classification, not
the map. Rust's and Julia's conformance recorders derived the corpus `role` from
the map name and are updated to defer to `observed_variables`.

**The consumer half (Python).** `simulation_array` builds the integrated `u`
layout from `state_variables`; a dual member would get a slot nothing advances
and read `0.0` out of it — which is #232's symptom 1 exactly, and its regression
tests caught it. The integrated set is now `state_variables` minus the names
`observed_variables` also holds: the defining equation writes that buffer.

## What I regenerated

**`tests/conformance/flatten/cases.json`** — regenerated with
`scripts/generate-flatten-corpus.py` **after** Python was fixed and verified by
hand against §4.7.5, never before. The diff is 2 lines of `$comment` plus exactly
one case:

* `edge_enumeration_area_eff` — its `edge_exists` and `edge_dense_id` are arrayed
  observeds written with the bare `index(V, sym)` LHS. They gain
  `observed_variables` membership and their `role` moves `state` → `observed`.

No equation, no ordering, no other case, and no other field moved. That the other
24 cases are byte-identical is the evidence that nothing collateral changed.

Nothing else is generated. The `classification_indexed_lhs` golden's new
`flatten_buckets` section is **authored** against the §4.7.5 table, not minted by
a binding.

## Conformance coverage added

1. **`classification_indexed_lhs`'s golden gains an optional `flatten_buckets`
   section** (`state_variables` / `observed_variables` / `algebraic_variables`,
   dot-namespaced, in document order), and **all five readers compare it**. This
   is what pins the `aggregate{k}(index(w, k))` spelling; see the caveat below
   for why the shared flatten corpus cannot.
2. **Julia's flatten driver un-skips the case.** `_FC_DIVERGENCES` listed
   `state_variables` / `observed_variables` on `edge_enumeration_area_eff` with a
   comment reading "The §4.7.5 step-4 table says BOTH … the corpus cannot [pin
   it], because the oracle does not implement the observed half. Neither binding
   implements the table's full reading today." Both do now, so the skip is gone
   and the case is compared in full — a second, independent cross-binding pin.
3. **The corpus generator gains a self-assertion** that a name appearing in both
   maps is ARRAYED, since §4.7.5 keeps a *scalar* observed out of
   `state_variables`.

### The revert check

Each binding's source fix was reverted in isolation and the tests re-run:

| Binding | `flatten_buckets` test | shared flatten corpus |
|---|---|---|
| Python | FAIL | FAIL |
| Julia | FAIL | FAIL |
| Rust | FAIL | FAIL |
| Go | FAIL | FAIL |
| TypeScript | FAIL | FAIL (3 assertions) |

## Merged with `main`

Merged `origin/main` at `e88bee218` (37 commits: #244 per-field tolerance, #251
right-hand-side `D` chain rule, #278 affine state gather, plus the `rust-tests`
timeout raise). The merge was textually clean, and the semantic checks that
matter here both pass:

* **`tests/conformance/flatten/cases.json` regenerates byte-identically on the
  merged tree.** #251 changes what `flatten` emits, so this is the check that
  proves my corpus regeneration and main's flatten changes do not collide.
* Neither completeness pin moved on main (Rust `diagnostic.rs`, Julia
  `_KNOWN_DIAGNOSTIC_CODES`), and main touched none of `flatten.go`,
  `flatten.ts`, `namespacing.jl`, `classification.jl` or the corpus generator.
  It did touch `flatten.py` and `flatten.rs`; my hunks survive intact and the
  full suites below were run on the MERGED tree, not the pre-merge branch.

## Test evidence (all on the merged tree)

| Suite | Result |
|---|---|
| `cargo test --no-fail-fast` | **1538 passed, 0 failed**; `cargo fmt --check`, `cargo clippy --all-targets` clean |
| `go test ./...` | **all packages ok**, `go vet ./...` clean |
| `npm test` | **2683 passed, 76 files**; `tsc --noEmit` and prettier clean |
| `pytest tests/` | **3408 passed**, 19 failed + 3 errors — *identical to the pre-change baseline* on this box (`test_esio_provider`, `test_loader_ingest_and_select`, `test_loader_unit_conversion_conformance`; missing optional loader deps, unrelated) |
| Python lint | `ruff check src/ tests/ ../../scripts/` and `ruff format --check` — the CI job's exact scope — clean |
| Julia | `Pkg.test()` hangs on this box (shared depot lock), so a minimal dev-linked env ran the affected files: `flatten_conformance_test.jl` **2651 pass**, `flatten_test.jl` **185 pass**, `classification_test.jl` **206 pass**, plus `flattened_to_esm_test.jl`, `conformance_round_trip_test.jl`, `round_trip_regression_test.jl`, `graph*`, `shape_promotion*`, `observed_*`, `inline_tests`, `cadence`, `aggregate_conformance`, `arrayed_vars`, `join_namespacing`, `reference_graph`, `validate` — **all green** |

## One deviation, with evidence

#270 and the fixture README note that Python is self-inconsistent because
`flatten._normalize_indexed_observed_lhs` normalizes only the `aggregate`
spelling, not the bare `index(w, i)` one. **I did not widen that rewrite**, for a
concrete reason: it rewrites the flattened `equations` list, which the shared
corpus compares across all five bindings, and **no other binding rewrites it**.
Widening it changes `edge_enumeration_area_eff`'s first two equation LHSs from
`EdgeEnumerationAreaEff.edge_exists[EdgeEnumerationAreaEff.edge]` to
`EdgeEnumerationAreaEff.edge_exists` in Python alone — I measured this — which
would move the oracle away from the four bindings that currently match it.

Nothing needs the rewrite: `wb` is filed correctly now because the bucket
decision is read from the §6.3.1 **classification**, which is spelling-agnostic
and already correct in all five bindings as of #268/#276. The docstring records
the constraint so the next person does not re-try it.

That Python's flatten mutates definition equations no other binding mutates is a
real cross-binding divergence, but it is a separate defect from this one and is
not fixed here.

Refs #231, #232, #237, #250, #268, #272. Not marked as closing #270 — triage is
yours.

https://claude.ai/code/session_01B1oPxGtgFJMT46aMZg5xTH


ctessum-claude:
## CI note: `rust-tests (stable)` is a 45-minute job timeout, not a test failure

Every other job is green — all three Julia legs, all four Python legs, all three Go legs, both TypeScript legs, and `rust-tests (beta)` + `rust-tests (1.89.0)` (MSRV).

`rust-tests (stable)` was cancelled **twice** (original + `gh run rerun --failed`) at the same place. Its step list both times:

```
success    Build Rust library
success    Build CLI binary
success    Build WASM
success    Run WASM test suite
success    Run Rust tests      <-- the step that validates this PR
cancelled  Build bench targets
skipped    Generate Rust documentation
skipped    Post Cache Rust dependencies
```

The job ran 45m19s against `timeout-minutes: 45` (workflow line 354). On `main` the same job takes ~39 minutes, so there is ~6 minutes of headroom and this leg went over it; `beta` and `1.89.0` took ~24-30 min on this same branch.

It is self-perpetuating on a branch once it trips: the cancellation skips `Post Cache Rust dependencies`, so the cargo cache is never saved and the next attempt starts just as cold.

Locally, `cargo test -j 4 --no-fail-fast` on this branch is **1517 passed, 0 failed**, with `cargo fmt --check` and `cargo clippy --all-targets` clean.

Consequence worth flagging: `standard-conformance-testing` is `skipped` because it depends on every language job, so **the cross-language gate did not run on this PR**. Raising `timeout-minutes` for `rust-tests` would fix both, but that is a workflow change outside this PR's scope and your call.

PR #292, opened 2026-09-09
perf(oop): the scalar read surfaces DO hold the last two transport scatters -- and freeing them costs 4.3% (default OFF; stacks on #283)

ctessum-claude:
**This arm ships DEFAULT OFF because it measured SLOWER.** It is correct, its
coverage is real, and it closes `SSA_SPIKE.md` blocker #4 — and on the CONUS 4x5
transport reverse it costs **4.3%**. `ESS_OOP_SSA_SCALAR=1` opts in; nothing on
the default path moves.

The durable parts of this PR are the blocker-tally instrumentation, the 108
assertions with their negative controls, and the measurement that says why the
arm is off.

**Stacks on #283** (`ctessum-claude:perf/ssa-read-redirect-arms`). GitHub will
not take a base branch that lives on the fork, so this PR is opened against
`main` and its diff therefore CONTAINS #283's three commits. **Review only the
last commit**; merge #283 first and this rebases to that one commit.

## The measurement

CONUS 4x5 (13x7x72), one process, both arm orders, `off` = #283 exactly,
`on` = #283 + this arm:

| program | off | on | ratio |
|---|---|---|---|
| `rhs` | 6.383 ms | 6.321 ms | 1.010 (flat) |
| `rhs_vjp` | 71.410 ms | 74.594 ms | **0.957 — 4.3% slower** |

The copy census (`p11_ue_traffic.py`, off the buffer-assignment dump) explains
it and is the reason the timing A/B was not run until it returned:

| | off | on |
|---|---|---|
| whole-buffer copies | 82 | **80** (−2) |
| total copy instructions | 290 | **432** (+142) |
| real element writes | 55.22 M | 56.11 M |

Two whole-buffer copies removed — exactly the two scatters the coverage number
says were freed — against 142 copy instructions added.

## Why it does not pay

The scalar surfaces' reads were **already cheap relative to the buffer versions
they kept alive**. On the transport RHS they are a handful of gathers per fill
level, not O(cells) of them; what made them matter to the *coverage* metric is
that a cheap read is still a read, and one read anywhere in a producer's block
pins that producer's whole scatter. So redirecting them frees two scatters —
real, and worth 2 whole-buffer copies — while the redirect's own materialization
adds cost: a producer-value gather per lane group, off a value XLA must now keep
live across the level boundary instead of letting it die into `ue`.

That is the same mechanism the concurrent scatter-skip **gate** work
established, not a separate mystery: **a partial skip pays twice** — once for
the buffer that still gets assembled by the producers that did not skip, and
once for the values kept alive beside it. Coverage is not the objective
function; the copy census is. Seven documented conclusions in this campaign
have been refuted by measurement and this is the eighth, on my own change.

## What the arm actually does (for review, and for whoever opts in)

One shared slot→(producer, position) map — the very vectors the plan build
already computes for the descriptor arm — published **per read surface as a
consumer LEVEL** (`_OopSSAOwn` + `_OopSSASCtx`). A scalar read of slot `s` at
level `ℓ` redirects iff `s` has an owner and that owner's level is strictly
below `ℓ`: the same two soundness facts the descriptor arm rests on (fills read
strictly lower levels; slot ownership is LAST-writer, so a slot a later kernel
rewrites is owned by that later kernel or by nobody and the level test refuses
it).

Not one slot-indexed table per surface: that would be O(levels × n_total) host
data (≈12 MB at CONUS) for a decision that is the same function of the slot
everywhere.

Granularity differs between the two walkers because their fallbacks do:

* `_oop_eval` reads ONE slot and can fall back per read ⇒ the verdict is per
  **slot**;
* `_oop_eval_batch` returns a whole lane vector through one gather ⇒ the verdict
  is per **node** and requires one common producer across every lane. Ghost
  lanes are wildcards (the caller's select overwrites them with 0), so a
  boundary lane does not break the single-producer form.

Build-time marking (`_ssa_mark_node_reads!` / `_ssa_mark_batch_reads!` /
`_ssa_mark_surface_reads!`) and the walk-time seams share **one** predicate,
`_ssa_owner_pid`, so a slot is marked residual exactly when the walker will read
it off `ue` — the two cannot drift. The feature's existing runtime guard still
covers the one divergence: a producer whose spine hoisted to a lane-invariant
scalar records no value, every redirect to it falls back to the (still-written)
gather, and `_oop_run_acc_vec` emits its scatter after all.

## Coverage, when opted in — ReSEACT transport RHS 6x6x8, `oop_ssa_stats`

| | arm off (= #283, = the shipped default) | arm on |
|---|---|---|
| scatters skipped | 13/17 (4 874 slots) | **15/17 (5 450 slots)** |
| `blockers` | scalar **2**, scan 1, sub 2 | scalar **0**, scan 1, sub 2 |
| `blockers_only` | scalar **2**, sub 1 | scalar **0**, sub 1 |
| scalar sites | 0/6 | **6/6** |
| scalar elements | 0 / 404 352 | **404 352 / 404 352** |
| descriptor edges | 97/100 | 97/100 |
| sub-kernel edges | 731/885 | 731/885 |

**Design-note item 4 was correct**: the surface is redirectable through the
existing map, and #283's per-reason tally is what pointed at it. It said
`blockers_only.scalar == 2`; a probe (not inference) then put both of those
producers on level-3/4 per-column fills reading a *lower* level's kernel output
through a `_NK_STATE_GATHER`, each batched into one lane group. Every such node
was "pure" — one common redirectable producer over its whole static slot table —
which is why the single-producer criterion suffices and why exactly those two
producers were released. That tally is also what let the concurrent gate work
find its own discriminator, which is the argument for keeping it.

With the knob off, **every** column of `oop_ssa_stats` reproduces #283 exactly
at that scale — only the new tally columns are populated — and the tests assert
that the shipped default is that control column for column.

The note's four host fixtures are unchanged (they were already 100% and have no
scalar read surface at all), which is why this branch adds fixtures that do.

## Exactness

`:oop`(arm on) == `:oop`(arm off) == `:oop`(shipped default) ==
`:oop`(feature off) == `:inplace` at Float64 `==`, and ForwardDiff jacobians
`==`, on all fixtures.

## Tests

`test/tree_walk_oop_ssa_test.jl` gains **108 assertions** in 4 testsets: the
redirect predicate and the lane-vector resolution as units (build-time verdict
asserted beside the walk-time one, because a disagreement is exactly what would
let a scatter be skipped under a live reader), and three end-to-end fixtures — a
leftover single on `_oop_eval`, a lane-batched `_NK_STATE` slot vector, and the
ReSEACT shape: a per-column fill whose contraction loop reads a level-1 kernel
producer through a `_NK_STATE_GATHER`. Every engagement assertion is paired with
the `ESS_OOP_SSA_SCALAR=0` negative control **and** with a shipped-default
assertion that `ESS_OOP_SSA=1` alone equals that control.

Host suite, 10 files, run with the emitter flag OFF and forced ON, both arms
identical and green: `tree_walk_oop` 164, `oop_scalar_batch` 42,
`array_obs_materialize` 71+3, `scan_prefix` 333, `observed_materialization`
8+12, `oop_merge`, `contraction_loop` 14+35+19, `tree_walk_observed_slots` 36,
`tree_walk_iip_generic` 69 (which also pins the in-place path's
zero-allocation property — threading the extra surface argument through the two
`:oop` walkers did not touch it), `tree_walk_oop_ssa` 274.

**Missing:** the full `Pkg.test()` sweep. That job died **OUT_OF_MEMORY at 64 G
after 54 minutes**; it needs `--mem>=128G` if anyone reruns it. It is moot for
merge now that the arm defaults off and the ten-file sweep above was green in
both arms.

https://claude.ai/code/session_01P67eFnfXX9S1XKTEiG2yd3

### DIFF
--- a/pkg/EarthSciAST.jl/src/tree_walk/SSA_SPIKE.md
+++ b/pkg/EarthSciAST.jl/src/tree_walk/SSA_SPIKE.md
@@ -180,14 +180,58 @@ fold.
   forcing reads go through the buffers argument, never `ue` — and the
   discrete-cadence refresh stays visible with fill scatters skipped (probed).
 
+## ReSEACT at CONUS: 1.36x on the whole adjoint loop, chemistry exact
+
+Measured 2026-09-09 at 4x5 CONUS (13x7x72, 6 552 cells), two driver builds in
+ONE process, arms interleaved, every pair run in BOTH arm orders — reproducing
+to three digits, so the order is not what is being measured.
+`tools/diag/p13_ssa_ab.jl` + `tools/diag/p11_ue_traffic.py` in the consuming
+repo.
+
+| program | arms off | arms on | off/on | copies | real element writes |
+|---|---|---|---|---|---|
+| `ros_step` (chemistry) | 38.90 ms | 28.23 ms | **1.38** (bit-for-bit) | | |
+| `ros_vjp` | 81.11 ms | 56.71 ms | **1.43** (λ bit-for-bit) | | |
+| `rhs` (transport RHS) | 3.755 ms | 6.405 ms | 0.59 | 2 → 3 | 2.49 M → 2.69 M |
+| `rhs_vjp` | 80.5 ms | 72.2 ms | **1.12** | **98 → 82** | 59.55 M → 55.22 M |
+| `ssp_step` (4-stage) | 13.41 ms | 24.25 ms | 0.55 | 8 → 12 | 10.43 M → 16.97 M |
+| `ssp_vjp` | 306.6 ms | 254.9 ms | **1.20** | **394 → 331** | 231.8 M → 176.3 M |
+
+Weighted by the adjoint's step mix (45.3 chemistry + 3.2 transport steps per
+300 s window) the per-window cost is 6.46 s → 4.74 s, **1.36x**.
+
+WHY THE TRANSPORT FORWARD LOSES. A producer's `dynamic_update_slice` into `ue`
+ALIASES its operand in the FORWARD, so XLA:CPU writes only the update and the
+scatter is nearly free there; skipping it instead forces the producer's value to
+be materialized as its own buffer. On the PPM transport RHS (17 producers,
+blocks up to 3 456 elements) that trade is a loss. On chemistry it is a win in
+both directions, because there the redirect replaces fragmented gathers of the
+12 326-element buffer with gathers of much smaller producer values and 59 of 59
+scatters disappear.
+
+The copies the scatter chain causes are all in the REVERSE, where each `ue`
+version has ~70 live consumers and copy-insertion duplicates the whole buffer
+per version. That is where removing readers pays on both halves: `ssp_vjp`
+loses 63 of its 394 whole-buffer copies and 24% of its write traffic for 1.20x,
+against a ~1.8x ceiling if every copy vanished.
+
+It is also why `ess-oop-levelbase` (one read version per level) failed where
+this succeeds: it reduced VERSIONS, which the forward already got for free,
+instead of removing READERS. Its census went 99 → 102 copies; this one goes
+98 → 82 and 394 → 331.
+
 ## What is NOT verified
 
-The `ESM_TEST_REACTANT=1` traced census (`reactant_oop_ssa_test.jl`) asserts the
-DUS delta equals `n_skipped_scatters` on the toy fixtures; the extension's arms
-are exercised there only insofar as those fixtures engage them. The ReSEACT
-timing A/B and the whole-buffer copy census live in the consuming repo
-(`tools/diag/p13_ssa_ab.jl`, `tools/diag/p11_ue_traffic.py`) because they need
-the offline forcing environment.
+Bit-identity across the arms is asserted and holds on HOST (`==`) and, on the
+toy fixtures, in the traced census (`reactant_oop_ssa_test.jl`, which pins the
+DUS delta to `n_skipped_scatters`). At CONUS the two arms are NOT bit-identical:
+changing the operand graph changes XLA's fusion, hence FMA and vectorization
+order. chemistry `ros_step` is bit-for-bit and `ros_vjp`'s λ is bit-for-bit (its
+p-gradient differs in the 20th significant figure); transport `ssp_step`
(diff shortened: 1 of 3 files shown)


ctessum-claude:
Closing this without merging. The work is right and the measurement is the
useful part of it; the code is not worth carrying.

**Why not merge.** Three things, reviewing the arm against today's `main`
rather than its own branch point:

1. **It is inert under the gate in #294.** `ESS_OOP_SSA_SKIP_WHOLE` (default on
   inside the feature) skips nothing unless *every* tracked producer's scatter
   can go: `all_skippable = n_skippable == length(prod_level) - 1`. This arm
   takes transport from 13/17 to 15/17, and the scan and sub-kernel blockers
   survive. 15 is not 17, so `build_ok` is false and zero scatters skip. With
   both merged at their defaults, this arm changes no emitted program. It only
   does anything under `SKIP_WHOLE=0`, which is the configuration #294 exists
   to refute.
2. **It measured slower** — 4.3% on the transport reverse, as this PR says
   plainly on its own change. That is a real result, not a reason to keep ~600
   lines of permanently-dead code implementing it.
3. **It conflicts with #294** in both `oop.jl` and
   `tree_walk_oop_ssa_test.jl`, so only one of the two could land as-is anyway.

None of that is a criticism of the PR. Shipping an arm default-off with the
census that says why, and calling it "the eighth refuted conclusion in this
campaign, on my own change", is exactly the right handling of a negative
result. The disagreement is only about whether the *implementation* should
live in the tree afterwards.

**What was kept.** The finding is now recorded in `SSA_SPIKE.md`, on blocker
#4 itself, which previously read as an open invitation to build this ("
redirectable through the same map; not done in the spike"). It now says the arm
was built, what it bought (13/17 -> 15/17, `blockers.scalar` -> 0), what it
cost (4.3%; two whole-buffer copies off against ~140 copy instructions added,
because these reads were already cheap relative to the buffer versions they
pinned), that it is inert under the shipped gate, and that it would only pay as
the *last* of the transport blockers to fall. So the next person reading that
list will not rebuild it from scratch.

The branch stays; reopen against a tree where the scan and sub-kernel blockers
are also closed, since that is the only configuration where this arm can pay.

PR #294, opened 2026-09-10
perf(oop): block size was the wrong gate -- the question is whether ue can die

ctessum-claude:
## What this is

**Stacks on #283** (`perf/ssa-read-redirect-arms`). GitHub will not take a
fork-hosted base branch, so this is opened against `main` and carries #283's
three commits underneath — **review only the last three commits**
(`3cc30d3bd`, `12ebd6799`, `84c7989e5`); everything before them is #283.

#283 made the `ess-oop-ssa` scatter-skip pay on chemistry and cost on
transport, and its own recommendation said the emitter "could gate the
scatter-skip on producer block size". **It measured, and block size is the
wrong quantity.** This adds the gate, on the quantity that turns out to
discriminate, and the loop goes 1.19x -> 1.26x.

## The verdict was STATIC; it needed to be a COST

The spike skipped a producer's `dynamic_update_slice` into the flat extended
buffer `ue` whenever static accounting showed nothing reads its block off the
buffer any more. Redundant is not the same as expensive: a DUS ALIASES its
operand in the forward, so XLA:CPU writes only the update and can fuse the
producer's spine straight into `ue`'s memory. Dropping it takes that away.

## The discriminator, and what it is not

`oop_ssa_producers(f)` is new reflection: per tracked producer, its fill level,
value length, how many redirected read surfaces now source from it and at what
element volume, how many take the whole value with no op at all, the
residual-read reasons, and both verdicts. On ReSEACT at 6x6x8 it says:

| | transport (part 1) | chemistry (part 2) |
|---|---|---|
| producers | 17 | 59 |
| statically skippable | 13 | **59 — all of them** |
| largest skipped block | 3 456 | **20 736** (then 10 368, 5 184) |
| redirected read volume / block | 0.125 … 58 230 | 1.0 … 198 |
| tier-1 whole-value reads | 4 producers, 1 each | 30+ producers, up to 87 |

**Block size is refuted**: chemistry skips producers 6x larger than
transport's largest and WINS. The read-volume ratio spans both extremes on
transport alone, so it does not separate the halves either.

What differs is **whether the flat buffer can DIE**. Chemistry skips 59 of 59,
so nothing scatters into `ue` and the whole buffer (12 326 elements at 6x6x8,
247 423 at CONUS) is dead code. Transport skips 13 of 17, so `ue` is assembled
anyway — and a PARTIAL skip pays twice: the four survivors still write the
buffer, AND each skipped producer's value becomes a buffer of its own instead
of fusing into an aliasing DUS.

### The test that could have refuted this

"A partial skip pays twice" and "one big producer is the problem" both predict
the regression; they differ on what dropping ONE skip does.
`ESS_OOP_SSA_SKIP_PIDS=!17` drops the skip of the largest producer (3 456
elements at 6x6x8, 78 624 at CONUS — 15x the next) and keeps the other twelve:

| `ssp_step` at 4x5 CONUS | flag off | #283 (13 of 17) | `!17` (12 of 17) |
|---|---|---|---|
| | 14.433 ms | 24.397 ms (0.592) | 24.768 ms (**0.583**) |

`no17` is not better than `ungated`. Size is not what makes a skip expensive;
partialness is. That is why the gate is all-or-nothing and why the two
per-producer bounds ship RELEASED — they are the refuted hypotheses, kept
bisectable.

## What changed

```
ESS_OOP_SSA_SKIP_WHOLE=1   (default) skip only when EVERY tracked producer's
                           scatter can go, so `ue` actually retires
ESS_OOP_SSA_SKIP_WHOLE=0   allow partial skipping (what #283 shipped)
ESS_OOP_SSA_SKIP=0         never skip — the negative control for the gate
ESS_OOP_SSA_SKIP_MAXLEN=n  per-producer block-size bound   (RELEASED by
ESS_OOP_SSA_SKIP_MINRATIO=x per-producer read-volume bound   default)
ESS_OOP_SSA_SKIP_PIDS=3,7  per-producer bisect; `!3,7` inverts
```

Following #283's pattern: every knob rides inside `ESS_OOP_SSA=1`, so one
process can A/B an arm and every engagement test has a negative control.

The verdict is now two passes — the static one, then the gate — and declining
a skip is always CORRECT: it emits a write nothing reads, exactly what the
flag-off build does. The gate touches ONLY whether the scatter is emitted; the
redirect tables are untouched, so `n_skippable_scatters` (a read-graph fact) is
identical in every arm and only `n_skipped_scatters` moves, with
`n_gate_declined` the difference. The redirect arms are not touched at all.

On ReSEACT the default gate leaves **chemistry structurally identical to #283**
(59 of 59 skipped, `n_gate_declined = 0`, edges 563/563) and takes **transport
to 0 of 17** with `n_skippable_scatters` still 13 and coverage unchanged
(97/100 top-level, 731/885 sub-kernel) — the redirects without the skip.

## Measured at 4x5 CONUS

13x7x72, 6 552 cells; three arms in ONE process, interleaved, both arm orders.
`noskip` is what the gate emits on transport, `ungated` is #283.

| program | flag off | gate | #283 |
|---|---|---|---|
| `ssp_step` (4-stage transport) | 14.668 ms | 19.887 ms (**0.738**) | 21.310 ms (0.688) |
| `rhs` (transport RHS) | 4.127 ms | 4.938 ms (**0.836**) | 6.523 ms (0.633) |
| `ssp_vjp` | 360.279 ms | 251.832 ms (**1.431**) | (1.203) |
| `ssp_vjp`, arms reversed | 347.620 ms | 254.526 ms (**1.366**) | |
| `ros_step` (chemistry) | | 1.38 | 1.38 |
| `ros_vjp` (chemistry) | | 1.43 | 1.43 |

The gate improves **both** sides of the transport trade: it roughly halves the
primal regression and it RAISES the VJP from 1.203 to 1.37-1.43, because the
redirects pay better once a partial skip is not working against them.

### The loop, counting replay

The adjoint evaluates the primal TWICE per accepted step — the forward
`T.step` and the backward `T.replay` — and the VJP once, so a primal
regression is paid twice against one VJP win. A per-window step+vjp mix
(#283's 1.36x) does not price that. Against the measured 2x2.5 48 h
decomposition (576 windows; transport step+replay 427 s, transport vjp
2 048 s):

| term | flag off | #283 | + this gate |
|---|---|---|---|
| chemistry (step + replay + vjp) | 5 106 s | 3 633 s | 3 633 s |
| transport (step + replay + vjp) | 2 475 s | 2 411 s | **1 995 s** |
| refresh | 1 792 s | 1 792 s | 1 792 s |
| host remainder | 123 s | 123 s | 123 s |
| **loop** | **9 495 s** | 7 959 s (1.19x) | **7 543 s (1.26x)** |

Transport was a WASH under #283 — the 1.20x VJP win almost exactly cancelled
the doubled primal regression, so the entire 1.19x was chemistry. With the gate
transport is a real 1.24x. 1.257x was PREDICTED from the break-even arithmetic
before the VJP was measured; 1.26x came back.

## The criterion NOT met

The bar for this work was the transport primal back to **>= 0.95** of flag-off.
It is **0.738** on `ssp_step` and 0.836 on `rhs`. The gate halves the
regression; it does not remove it, and 0.738 should not be read as success.

What makes it net-positive anyway is arithmetic, not judgement: the gate is
worth it iff `427 + 2048/r < 2411`, i.e. `ssp_vjp >= 1.032`. It delivers ~1.40.
So the trade is favourable by a wide margin rather than marginally — but a
caller that runs the transport primal with NO reverse at all is still 1.36x
worse off than flag-off (down from #283's 1.8x, not gone).

**Further work, not a claim**: with zero scatters skipped the only remaining
difference from flag-off is the redirects themselves, so something on the
transport primal still materializes that the flat-buffer gather did not — the
producer values the sub-kernel and tier-2b redirects read from. Pricing that
needs a redirect gate, which is a separate decision.

## Exactness

Unchanged, and it is the contract. Declining a skip only emits a write nothing
reads, so no arm can move a number:

* `:oop`(any gate arm) == `:oop`(off) == `:inplace` at Float64 `==`, and
  ForwardDiff jacobians `==`, on every fixture — asserted across all seven
  arms, not just the default.
* With the flag OFF every structure is the empty singleton and the emitted
  program is byte-for-byte the pre-spike emitter.
* Traced at CONUS the gate can only move transport CLOSER to flag-off, since it
  removes emitter changes rather than adding them; #283's figures (`ros_step`
  and `ros_vjp` λ bit-for-bit, `ssp_step` 5.6e-15 absolute = 1.1e-18 of scale,
  not one of 85 176 components over 1e-9 of scale) therefore bound this build.
  Note that the POINTWISE relative metric reads a spurious `2.0` on
  opposite-signed components at ~1e-310, which is why the probes report
  absolute and SCALE-relative figures plus a component count.

## Tests

141 new assertions in `test/tree_walk_oop_ssa_test.jl`, every knob with a
negative control:

* **the gate moves ONLY the skip verdict** — across all seven arms,
  `n_skippable_scatters`, `n_fast`, `elems_fast` and `n_sub_fast` are pinned
  equal and `n_skipped + n_gate_declined` is invariant;
* **each knob declines what it says it declines** — ungated == 2, master off
  == 0, `MAXLEN=1` == 0, `MINRATIO=1e9` == 0, `PIDS=!2` == 1 *and it is
  producer 2's scatter that survived*, and the shipped default is not
  degenerate on the fixtures;
* **the all-or-nothing cascade** on a new fixture (`_s_mixed`) whose second
  producer can be blocked ON DEMAND: `ESS_OOP_SSA_PGATHER=0` leaves the build
  only partially skippable, the whole gate then declines every skip
  (`n_gate_declined == 1` with `n_skippable_scatters` unmoved at 1), and
  `ESS_OOP_SSA_SKIP_WHOLE=0` restores exactly the producer the static verdict
  named;
* **the producer table's columns** against shapes whose read structure is known
  by construction — the fan-out's two 7-element producers at `nread == 4`,
  `nwhole == 4`, `elread == 28`, and the merged class's ONE 12-slot producer
  read by two 6-element slices at `nwhole == 0`, `nread == 2`, `elread == 12`;
* **values and derivatives in every arm**, `==` against flag-off and
  `:inplace`.

`tree_walk_oop_ssa_test.jl`: 45 + 88 + 16 + 17 + 92 + 16 + 33, all green.

## Recommendation on defaulting `ESS_OOP_SSA`

#283 recommended defaulting it with the caveat that "the transport forward is
1.8x slower, so a caller that runs the transport primal without a reverse is
worse off". **This gate reduces that caveat but does not remove it**: such a
caller is 1.36x worse off, not 1.8x. So the honest answer is:

* **for an adjoint / gradient workload, yes** — the loop is 1.26x cheaper
  counting replay, both halves now win in both directions, and chemistry is
  exact;
* **the caveat stands in weaker form** for a primal-only transport caller, and
  the second of #283's caveats stands unchanged: flag-on is not bit-identical
  to flag-off in the traced transport path (5.6e-15, 1.1e-18 of scale), so
  defaulting it changes numbers for existing traced users and should be a noted
  change, not a silent one.

I would default `ESS_OOP_SSA=1` and note both, rather than default it silently.

## Not verified here

The ReSEACT timings live in the consuming repo (`tools/diag/p14_ssa_gate.jl`,
the N-arm successor to `p13_ssa_ab.jl`, plus `p14_gate_primal.sbatch`); they
need the offline forcing environment. The whole-extended-buffer copy census was
not re-run for the gate — the timing three-arm A/B was the priority.

### DIFF
--- a/pkg/EarthSciAST.jl/src/tree_walk/SSA_SPIKE.md
+++ b/pkg/EarthSciAST.jl/src/tree_walk/SSA_SPIKE.md
@@ -180,14 +180,58 @@ fold.
   forcing reads go through the buffers argument, never `ue` — and the
   discrete-cadence refresh stays visible with fill scatters skipped (probed).
 
+## ReSEACT at CONUS: 1.36x on the whole adjoint loop, chemistry exact
+
+Measured 2026-09-09 at 4x5 CONUS (13x7x72, 6 552 cells), two driver builds in
+ONE process, arms interleaved, every pair run in BOTH arm orders — reproducing
+to three digits, so the order is not what is being measured.
+`tools/diag/p13_ssa_ab.jl` + `tools/diag/p11_ue_traffic.py` in the consuming
+repo.
+
+| program | arms off | arms on | off/on | copies | real element writes |
+|---|---|---|---|---|---|
+| `ros_step` (chemistry) | 38.90 ms | 28.23 ms | **1.38** (bit-for-bit) | | |
+| `ros_vjp` | 81.11 ms | 56.71 ms | **1.43** (λ bit-for-bit) | | |
+| `rhs` (transport RHS) | 3.755 ms | 6.405 ms | 0.59 | 2 → 3 | 2.49 M → 2.69 M |
+| `rhs_vjp` | 80.5 ms | 72.2 ms | **1.12** | **98 → 82** | 59.55 M → 55.22 M |
+| `ssp_step` (4-stage) | 13.41 ms | 24.25 ms | 0.55 | 8 → 12 | 10.43 M → 16.97 M |
+| `ssp_vjp` | 306.6 ms | 254.9 ms | **1.20** | **394 → 331** | 231.8 M → 176.3 M |
+
+Weighted by the adjoint's step mix (45.3 chemistry + 3.2 transport steps per
+300 s window) the per-window cost is 6.46 s → 4.74 s, **1.36x**.
+
+WHY THE TRANSPORT FORWARD LOSES. A producer's `dynamic_update_slice` into `ue`
+ALIASES its operand in the FORWARD, so XLA:CPU writes only the update and the
+scatter is nearly free there; skipping it instead forces the producer's value to
+be materialized as its own buffer. On the PPM transport RHS (17 producers,
+blocks up to 3 456 elements) that trade is a loss. On chemistry it is a win in
+both directions, because there the redirect replaces fragmented gathers of the
+12 326-element buffer with gathers of much smaller producer values and 59 of 59
+scatters disappear.
+
+The copies the scatter chain causes are all in the REVERSE, where each `ue`
+version has ~70 live consumers and copy-insertion duplicates the whole buffer
+per version. That is where removing readers pays on both halves: `ssp_vjp`
+loses 63 of its 394 whole-buffer copies and 24% of its write traffic for 1.20x,
+against a ~1.8x ceiling if every copy vanished.
+
+It is also why `ess-oop-levelbase` (one read version per level) failed where
+this succeeds: it reduced VERSIONS, which the forward already got for free,
+instead of removing READERS. Its census went 99 → 102 copies; this one goes
+98 → 82 and 394 → 331.
+
 ## What is NOT verified
 
-The `ESM_TEST_REACTANT=1` traced census (`reactant_oop_ssa_test.jl`) asserts the
-DUS delta equals `n_skipped_scatters` on the toy fixtures; the extension's arms
-are exercised there only insofar as those fixtures engage them. The ReSEACT
-timing A/B and the whole-buffer copy census live in the consuming repo
-(`tools/diag/p13_ssa_ab.jl`, `tools/diag/p11_ue_traffic.py`) because they need
-the offline forcing environment.
+Bit-identity across the arms is asserted and holds on HOST (`==`) and, on the
+toy fixtures, in the traced census (`reactant_oop_ssa_test.jl`, which pins the
+DUS delta to `n_skipped_scatters`). At CONUS the two arms are NOT bit-identical:
+changing the operand graph changes XLA's fusion, hence FMA and vectorization
+order. chemistry `ros_step` is bit-for-bit and `ros_vjp`'s λ is bit-for-bit (its
+p-gradient differs in the 20th significant figure); transport `ssp_step`
+differs by 5.6e-15 pointwise relative. The pointwise metric on the VJPs is
+uninformative — it reads exactly 2.0 on components at ~1e-310 — so the probe
+also reports a scale-relative figure and the count of components over 1e-9 of
+scale.
 
 ## Generalization assessment (honest)
 
--- a/pkg/EarthSciAST.jl/src/tree_walk/SSA_SPIKE.md
+++ b/pkg/EarthSciAST.jl/src/tree_walk/SSA_SPIKE.md
@@ -156,9 +156,21 @@ fold.
    repeat cells — median run length 1 — so slices+concat lose to the gather;
    would need segment-level ops instead. Left as gathers deliberately.
 4. **Scalar-walker and lane-batched scalar reads** (`_NK_STATE`,
-   `_NK_STATE_GATHER`, batch slot vectors). Redirectable through the same
-   slot→(producer, position) map; not done in the spike. They also hold
-   producer scatters alive wherever they read.
+   `_NK_STATE_GATHER`, batch slot vectors) — REDIRECTABLE, MEASURED, and NOT
+   WORTH IT. They are reachable through the same slot→(producer, position) map,
+   published per read surface as a consumer level; an implementation took
+   transport from 13/17 to 15/17 scatters skipped and `blockers.scalar` to 0.
+   It cost **4.3% on the transport reverse**: two whole-buffer copies removed
+   against ~140 copy instructions added, because these reads were already cheap
+   relative to the buffer versions they pinned, and redirecting one keeps a
+   producer value live across a level boundary instead of letting it die into
+   `ue`. Coverage is not the objective function; the copy census is.
+
+   It is also INERT under the shipped gate. `ESS_OOP_SSA_SKIP_WHOLE` skips
+   nothing unless EVERY tracked producer can go, and 15 of 17 is not 17 of 17 —
+   the scan and sub-kernel blockers below survive it. So closing #4 alone
+   changes no emitted program at the default; it would only pay as the LAST of
+   the transport blockers to fall. Do not re-attempt it on its own.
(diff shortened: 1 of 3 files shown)


ctessum-claude:
Reviewed as a group with #283 and #292 against current `main`. **Merging this,
straight after #283.** #292 is closed — see the note there.

The argument that carried it: #283 alone ships a design note whose headline
number this PR refutes, and recommends the ungated configuration this PR shows
is the worse of the two. They belong in the tree together or not at all.

The refuted hypotheses being kept *bisectable* rather than deleted —
`SKIP_MAXLEN` and `SKIP_MINRATIO` released by default, `SKIP_PIDS=!17` as the
test that could have falsified "a partial skip pays twice" — is most of why the
positive result here is believable.

## Two corrections pushed

**`8e9b3f83d`** — the knob-group comment said *"Both bounds default to the
measured gate"*, contradicted by the code four lines below it
(`_SSA_SKIP_MAXLEN_DEFAULT = typemax(Int)`, `_SSA_SKIP_MINRATIO_DEFAULT = 0.0`)
and by the very next comment, *"Both PER-PRODUCER bounds are released by
default"*. Stale prose from the block-size hypothesis those bounds exist to
test and that this PR refutes. Someone following it would set
`ESS_OOP_SSA_SKIP_GATE=0` expecting to lift a bound that was never applied. The
comment now says what that switch actually does — release both bounds and the
whole-buffer gate together. Comment-only; verified no non-comment line moved.

**`0c29c8adb`** — `SSA_SPIKE.md` blocker #4 still read as an open invitation
("redirectable through the same map; not done in the spike"). It *was* done, in
#292, and measured: 13/17 to 15/17, `blockers.scalar` to 0, at 4.3% on the
transport reverse. It is also inert under the gate this PR ships, since 15 of 17
is not 17 of 17. Recorded on the list itself so it is not rebuilt from it.

## What I checked

Gate soundness is the easy half and it holds: declining a skip only emits a
write nothing reads, so no arm can move a number, and the gate touches only
`n_skipped_scatters` while `n_skippable_scatters` stays a read-graph fact.

The real exposure is calibration, and I filed it as **#296**. `build_ok` turns
on a single integer comparison — `n_skippable == length(prod_level) - 1` — and
every count feeding it was measured before #278 landed, which moves state
gathers between the affine and tabled tiers and so changes the descriptor
population this plan resolves. If transport now reaches 17 of 17, the gate flips
from declining every transport skip to taking them all, which is the regime #283
measured at 1.8x *slower*. Not a soundness problem — I confirmed both descriptor
kinds are enumerated by the residual marking and that `_OopAccPlan` / `_AccDesc`
are byte-identical between the fork point and `main` — and the feature is
default-off. But it should be re-measured before anyone acts on the note's
recommendation to default `ESS_OOP_SSA=1`.

The honesty about the criterion not met (0.738 against a 0.95 bar, stated
outright rather than buried) is what makes the rest of the numbers here worth
trusting.

PR #299, opened 2026-09-11
feat(schema): rename the `aggregate` node to `faq` at esm 1.1.0

ctessum-claude:
## Why

`aggregate` names one specialization of the unified query node and mislabels the rest. The node also **joins** (`join`), **filters** (`filter`), and under `bool_and_or` with `distinct` is **index-set-producing** — it materializes a `kind: "derived"` index set rather than an array, reducing nothing at all. A reader who meets `"op": "aggregate"` and reasons from the English word forms a model that is wrong in every one of those cases.

It has been a **Functional Aggregate Query** since RFC `semiring-faq-unified-ir` introduced it — the name is already in that RFC's title, in `CONFORMANCE_SPEC.md`, and in the `area_faq` module in four of five bindings. The serialized tag was the last place still using the narrow word.

RFC: `docs/content/rfcs/faq-node-rename.md`.

## What changes

- **`"op": "faq"` is canonical** at esm 1.1.0 (unreleased — no new version minted).
- **`aggregate` is a deprecated alias.** Accepted on load, normalized to `faq` at the wire boundary — ahead of the version gates and schema validation, so nothing downstream of the loader (`emit` included) ever sees it — and reported once per document as the new warning `deprecated_op_alias`, naming the node count. **Removed at esm 2.0.0**, stated normatively in the schema and `esm-spec` §4.2, not just in the RFC.
- **`arrayop` is removed**, and rejected by name with the new hard error `removed_op`.
- **The text surface is unchanged.** The head keyword there is the ⊕-word (`sum`/`prod`/`max`/`min`/`any`) chosen from `semiring`/`reduce` — the ASCII spelling of `Σ`/`Π`/`⋁` in the unicode and LaTeX renderings — not the node's name. Parsers now emit `faq`; `RENDERING_CONTRACT.md` is untouched.

## Why `arrayop` needed a second removal

It was declared removed at 0.8.0. That change removed it from the **load** path but not the **construction** path: three Julia sites went on manufacturing `arrayop` nodes internally, and ~33 defensive `op == "aggregate" || op == "arrayop"` branches across **all five** bindings existed to keep them working. The tag stopped being a file-level alias and quietly became an *internal node kind* — never warned about, absent from the corpus, and absent from the schema conditional that requires `output_idx`/`expr` (so an `arrayop` node bypassed required-field validation entirely).

Hence the rule this PR adopts for its own alias: **a deprecation without a removal version is a rename that never finishes.**

The Julia ModelingToolkit extension keeps the name where it refers to SymbolicUtils' own `ArrayOp` construct (`ext/mtk_ext/arrayop.jl`, `_build_arrayop_sym`, the `@arrayop` comments) — upstream's vocabulary for an upstream type. Only its wire-tag dispatch moved.

## Two decisions worth reviewer attention

**1. `arrayop` is rejected BY NAME, not left to the open tier.** It is a well-formed identifier, so esm-spec §4.2 would otherwise admit it as an OPEN rewrite-target op: the document loads silently and fails much later as `unlowered_operator`, or never. Simply deleting the alias is the obvious-and-wrong implementation. New diagnostic `removed_op` (hard error); Julia's removed `call` op is the precedent.

**2. `faq` is version-gated at 1.1.0, transitively.** Same rule as the top-level `solver` block (`reject_solver_pre_v11`). The gate carries through *lowering*: a template-library consumer whose imported rules **expand** to `faq` declares 1.1.0 even though its authored bytes carry no `faq` node. This is what the conformance golden pairs needed — goldens and their inputs were bumped as units.

## Normalize on load, don't preserve on emit

`emit` writes `faq` for a document authored with `aggregate`. This is a deliberate, bounded dent in the 0.9.0 `emit ∘ load` byte-wise fixed point, restated as: the fixed point holds for documents **at the current schema version**, and a document carrying a deprecated alias is upgraded exactly once, on its first load.

The alternative — preserving the authored spelling through to emission — was rejected: it puts both spellings into all five emitters, and the round-trip conformance gate compares each binding against **its own** re-parse, so a divergence in which spelling a binding preserves would not be caught.

## Tests

New gate `tests/conformance/deprecated_op_alias/` — `aliased.esm` and `canonical.esm` differ **only** in the op tag, so "byte-identical after emit" is literally checkable. One test file per binding drives it, plus `tests/invalid/faq/arrayop_op_removed.esm` with its pin in `expected_errors.json`, and three new `SCHEMA-03-A-*` rows in the compliance matrix.

The corpus is migrated (634 tags, 195 version bumps) rather than left on the alias — otherwise the whole corpus would exercise the deprecated path while the canonical tag went barely tested.

| Binding | Result |
|---|---|
| TypeScript | 2713 passed / 0 failed (80 files) |
| Python | 3433 passed / 0 failed |
| Go | 2207 passed / 0 failed |
| Rust | 1554 passed / 0 failed (143 suites) |
| Julia | alias suite 7/7 |

**Not run:** the full cross-language `./scripts/test-conformance.sh`. Each binding's own suite is green, but the cross-binding comparison stage has not been exercised — worth a CI run before merge.

Pre-existing failures excluded from the Python count (files untouched by this PR, `earthsciio` dependency drift — the installed package exposes `DataLoader`, not `DataSource`): `test_esio_provider.py`, `test_loader_ingest_and_select.py`, `test_loader_unit_conversion_conformance.py`.

https://claude.ai/code/session_01JoQCxbmggBmHkUqABTjtQX


ctessum-claude:
## Review addressed

All three blocking findings were real, and all three survived five green binding suites. Fixed in `251c562e9`.

**1. Referenced documents bypassed the normalizer.** The rewrite ran on the root's bytes; `{ref}`/import resolution then parsed child files raw, so a child's `aggregate` reached `emit` and a child's `arrayop` loaded silently — proven by probe in Python and Rust, traced in Julia's remote-ref path. Each binding now has **one chokepoint** every document passes through (root, `{ref}` child, template library, coupling library). Julia's `_read_json_document` already called itself *"THE wire boundary"* in its own docstring while four sites bypassed it.

**2. MTK extension dispatched on a dead tag.** `ext/mtk_ext/arrayop.jl` tested `lhs.op == "arrayop" || "aggregate"`, neither of which can occur now, so the LHS grid-shape override silently became a no-op — wrong-sized arrays, no error. The MTK carve-out is for SymbolicUtils' `ArrayOp` vocabulary, not wire-tag dispatch.

**3. Version gate implemented** (`faq_version_too_old`). It reads the **authored** form deliberately: `aggregate` *is* the pre-1.1.0 spelling, so a 1.0.0 document carrying the alias stays legal and is normalized with its version raised to the floor.

### What turning the gate on exposed

~104 test failures, every one a real inconsistency the rename had created (ops rewritten to `faq`, declared versions left at 1.0.0 or 0.8.0).

It also exposed a gap in the design itself: a 1.0.0 parent that **mounts** a child using `faq` has the child inlined at load, so it comes to *contain* `faq` without ever spelling it — and emitting it as 1.0.0 wrote a document the gate then refused to read back. Failed identically in Rust and Python on `valid/subsystem_index_set_merge`. Emit now stamps a 1.1.0 floor, the same shape as the §9.6.4 rule-8 stamp already documented as *"a minimum that may only ever raise"*.

### Also fixed

Go's normalizer destroyed authored key order (decode + re-marshal sorts keys) — now an order-preserving textual rewrite; the unsatisfiable `removed_op` pin, re-categorized as `parse_error`; Julia's self-contradictory message; the TS double-warn; Python's invisible `DeprecationWarning`; the vacuous Rust test; the Python/TS caller-mutation divergence.

**Prose:** `semiring-faq-unified-ir.md` and the dated audit are **restored from main** rather than patched. That RFC's §5.6 is *about* the arrayop/aggregate distinction, so sweeping both tags turned it into tautologies ("accepts `faq` as an exact synonym for `faq`"); a dated audit is a record, not documentation. The RFC gets an amendment banner only. The schema and esm-spec also claimed `arrayop` is "rejected like any other unknown op" — it is rejected *by name*, which is the entire point.

### Why nothing caught this

Every fixture in the new conformance suite was a single self-contained document, and the suite is driven by five independent per-binding unit tests that are never compared against each other. It now mounts a child by `{ref}` in both the alias and `arrayop` shapes, plus version-gate cases.

### Verified

| Binding | Result |
|---|---|
| TypeScript | 2717 / 2717 |
| Rust | 1558 / 0 (143 suites) |
| Python | 3437 / 0 |
| Go | ok |
| Julia | alias suite 14/14 |

Still not run: the full cross-language `./scripts/test-conformance.sh`.

Issue #230, opened 2026-09-07
operator_compose's merged-away rename reaches equation ASTs only, not coupling endpoints in later entries or runner name lookups


ctessum:
## Symptom

`operator_compose`'s merged-away rename is **document-wide over equation ASTs and nothing else**. When a merge consumes `B.x` and retargets it onto `A.x`, every equation in the document is rewritten off the dead name — but three other places that address the state *by name* are not:

1. **Coupling endpoints in later entries.** `couple` connector `from`/`to`, `variable_map` `from`/`to`, and a later `operator_compose`'s `translate` endpoints are plain scoped-reference strings on the entry objects. The flattener applies `operator_compose` entries first (esm-libraries-spec §4.7.5 step 3 ordering), so an entry that runs afterwards can still name a spelling the merge has already deleted.
2. **Runner lookups keyed on the name.** `parameter_overrides` keys, output/plot variable selections, and any consumer holding `"SuperFast.O3"` as a string address a state that has moved.
3. **A surviving expression-template registry body** that references the dead name, which expands at the build boundary rather than at flatten.

The issue text of #195 called this out directly:

> The rename is also document-wide (`_retarget_merged_names`, `flatten.py:1324`; `coupling_apply.jl:452-459`) but reaches **equation ASTs only** — not coupling endpoints in later entries, not runner lookups keyed on the name. So the states move out from under everything downstream that addresses them by name.

## Why it is filed separately from #195

#195 is closed by #216, which (a) reports a zero-/partial-merge, (b) adds a tri-state `require_match`, and (c) refuses an ambiguous bare-name match rather than silently choosing which of two states survives.

(c) **narrows** this hazard substantially but does not remove it. A bare-name match between two states is now refused outright, so the accidental rename that motivated #195 cannot happen. What remains is the rename that is still legal and still silent:

- a `translate` match, where the author deliberately names the surviving spelling — `B.x` really is deleted, and a later `variable_map` naming `B.x` still dangles;
- a bare-name match where only one side is a state (the other an observed), which #216 resolves by ownership rather than refusing.

So the reach gap is now a smaller, sharper thing than it was: it applies to renames the author *asked for*, which is arguably where it matters most, because those are the ones a document is built around.

## What fixing it involves

The rename map has to be threaded out of the merge and applied to:

[...]

PR #315, opened 2026-09-13
fix(julia): a container's inline tests build over its DOCUMENT, not the container alone (#314)

### LINKED ISSUE #314: A container's inline tests build over the container, not its document: a document-scoped template registry and every sibling component are dropped (Julia)
## Summary

The Julia inline-test runner compiles each container from the container itself, not from its document. `run_file_tests!` (`pkg/EarthSciAST.jl/src/run_tests.jl`) calls `MTK.System(model::Model)`, which wraps that one model in a **synthetic single-model `EsmFile`** (`flatten(::Model)`, `src/flatten.jl:1313`) and flattens that. Everything the document supplies around the container is dropped, and two of those things the live corpus relies on:

**A — the document's `expression_templates` registry is lost.** `Model` has no `expression_templates` field; the registry is document-scoped. So the synthetic file carries an empty one, `expand_flattened_refs` (called at the top of `ModelingToolkit.System(::FlattenedSystem)`) returns early on `isempty(reg)`, and a §9.6.4 Option-B `apply_expression_template` reference — the form the spec says **survives** the load-time fixpoint precisely so the build can resolve it — reaches `_esm_to_symbolic` with no arm:

```
Model compile failed: ArgumentError("Unsupported operator: apply_expression_template")
```

**B — every sibling component's variables are lost.** An equation that reads another component of the same document (`D(Rs.A, t)`, the §6.6 instantaneous-derivative shape for a mechanism's species tendencies, or a plain `Rs.A`) dies in `_resolve_lowering_var`:

```
Model compile failed: ArgumentError("Variable 'Rs.A' not found in variable dictionary")
```

These are two faces of one bug, which is why they are one issue: **the container is not the document.**

## Root cause, probed

`flatten(file)` is already correct — it namespaces species and models into one registry. Only the per-container path is wrong:

```
model fields: (:variables, :equations, :discrete_events, :continuous_events, :subsystems,
               :tolerance, :tests, :analyses, :initialization_equations, :guesses,
               :system_kind, :reference)      # ← no expression_templates
FILE  flatten registry keys:  ["arrhenius"]   → expand_flattened_refs expands
MODEL flatten registry keys:  String[]        → expand_flattened_refs is a no-op
```

and for B, the same file's document flatten has exactly what the lowering says is missing:

```
state_variables: ["Rs.A", "Rs.B", "Rs.P"]
observed:        ["Reader.conc_A"]
```

## Cross-binding verdict

esm-spec §6.6 selects **which tests run**, not what the system contains. The other two executing runners both build over the whole document, so Julia is the 1-of-3 outlier:

| runner | build scope | A | B |
|---|---|---|---|
| Rust `esm test` | whole flattened document | pass | pass |
| Python gate (`flatten(ef)` → `esm_problem`, `tools/run_esm_inline_tests.py:47`) | whole flattened document | pass | pass |
| **Julia `run_esm_tests`** | **the container alone** | **`Unsupported operator`** | **`not found in variable dictionary`** |

Go and TypeScript execute no inline tests, so they have no position here.

## Repro

Three fixtures plus a control (`data/eqwefic/esm-repro/julia-gaps/cases/`, also added as `pkg/EarthSciAST.jl/test/fixtures/container_in_document/`):

| fixture | shape | Rust | Python | Julia |
|---|---|---|---|---|
| `surviving_template_ref.esm` | model equation RHS is a non-eager `apply_expression_template` | pass | pass | **A** |
| `sibling_species_read.esm` | model observed reads `Rs.A` directly | pass | pass | **B** |
| `sibling_species_tendency.esm` | model observeds are `D(Rs.<X>, t)` for each species | pass | pass | **B** |
| `sibling_species_system_alone.esm` (control) | the mechanism alone, asserting its own trajectory | pass | pass | pass |

The control is the point: a mechanism on its own is fine. It breaks only when a sibling component needs to see it.

Note for anyone reading the original report: the first `not found in variable dictionary` observed in the wild named `SULF`, a product-only species, which suggested product-only species were being dropped from the stoichiometry. They are not — `lower_reactions_to_equations` handles them correctly, and the minimal repro fails on `Rs.A`, an ordinary substrate. `SULF` was simply first in declaration order.

## Effect on the live corpus

In EarthSciML/EarthSciModels (run 34726972505), these two signatures account for whole files being reported as 0-pass:

- `components/gaschem/radm2/radm2.esm` — 0 pass / 115 err (RADM2RateConstants is A, RADM2Tendencies is B, and the mechanism's own trajectory tests then fail in the solve because their `parameter_overrides` name the sibling model's photolysis parameters)
- `components/urban_canopy/urban_radiation.esm` — 0 pass / 13 err

The Rust CLI runs the first at 825/825 and the second at 31/31.

ctessum-claude:
Fixes #314.

## The bug

`run_file_tests!` compiled each container through `MTK.System(model::Model)`, which wraps that one model in a **synthetic single-model `EsmFile`** (`flatten(::Model)`) and flattens that. Everything the document supplies around the container went with it:

- **The document-scoped `expression_templates` registry.** `Model` has no such field. The synthetic file carried an empty registry, so `expand_flattened_refs` — which `ModelingToolkit.System(::FlattenedSystem)` calls at its entry precisely to honour "expand at your boundary" — returned early on `isempty(reg)` and did nothing. A §9.6.4 Option-B `apply_expression_template` reference, the form the spec says **survives** the load-time fixpoint so the build can resolve it against the retained registry, then reached `_esm_to_symbolic` as `Unsupported operator: apply_expression_template`.
- **Every sibling component's variables.** An equation reading another component of the same document — `D(Rs.A, t)`, the §6.6 instantaneous-derivative shape for a mechanism's species tendencies, or a plain `Rs.A` — died in `_resolve_lowering_var` as `Variable 'Rs.A' not found in variable dictionary`.

One mistake, two faces: the container is not the document.

## Why this direction

`flatten(file)` was already right; only the per-container build was wrong. esm-spec §6.6 selects **which tests run**, not what the system contains, and the other two executing runners already build over the whole document — the Python gate does `flatten(ef)` then `esm_problem` (`tools/run_esm_inline_tests.py`), and the Rust CLI builds `esm_problem` over the whole flattened document. Julia was the 1-of-3 outlier.

## What changed

`_compile_model_in_document` builds from a per-file `FlattenedSystem`, computed **at most once** and **only when a container that needs it has tests** — a container with no tests is still never compiled, so §9.5.3a's refusal of an unimplementable table stays attached to a build that actually happens. Table lowering moves to file scope for the same reason (the flatten happens once, ahead of the per-container builds).

A mechanism sharing its document with a model takes the same path. RADM2's 156 reaction rates are scoped references into the component that evaluates its rate coefficients, and its tests override that component's parameters; the isolated Catalyst build cannot see it, so the species resolved and the **solve** failed rather than the build (`Variable 'RADM2RateConstants.j_ch3o2h' not found on compiled system`). A document whose **only** containers are reaction systems keeps the Catalyst route deliberately: that is the path carrying Catalyst's own default seeding, and the one the corpus's large mechanisms — including the `geoschem_fullchem` build this rig's CI comments call out as OOM-prone in-process — have always taken. This change does not move them.

## Evidence

**The new test fails 5 of its 16 assertions without the source change** (`7 passed, 5 failed`), and passes 16/16 with it. Fixtures: a surviving template reference, a sibling-species read, a sibling-species tendency, and — the one that says this is about the container/document boundary rather than about reaction systems — a control in which the same mechanism, alone in its document, passes before and after.

Measured against the live EarthSciML/EarthSciModels corpus, with the Rust CLI as the reference:

| file | Julia before | Julia after | Rust |
|---|---|---|---|
| `components/gaschem/radm2/radm2.esm` | 0 pass / 115 err | **825 / 0** | 825 / 0 |
| `components/urban_canopy/urban_radiation.esm` | 0 pass / 13 err | **31 / 0** | 31 / 0 |

Files failing for other reasons are unchanged — I checked `brox_cycle.esm` (52/2), `nox_cycle.esm` (67/0), `stratospheric_ozone_system.esm` (26/150) and `geoschem_ratelaws.esm` (36/1) against their pre-change CI rows and they are identical. Two further Julia gaps are visible in that corpus and are **not** addressed here: `Unsupported operator: and` (`fuel_model_lookup.esm`) and an unqualified `Variable 'T_ww' not found in variable dictionary` (`urban_canopy_model.esm`, a subsystem-scoped name rather than a sibling-component one).

## Test status

`Pkg.test()` was still running when this was opened; I will report the result on the PR rather than claim it here.

---

## Review pass (2026-09-13)

The direction is right — esm-spec §6.6 says a test declared on a component the
document itself declares "is built from the document it lives in, coupling and
all", and the Julia tree-walk runner (`run_inline_tests`), the Rust CLI and the
Python gate all already did that. Running the branch's runner over the
repository's own multi-component documents with inline tests found one document
this branch fixed that nobody had claimed — `tests/conformance/rhs_time_derivative/fixtures/tendency_resolution.esm`,
from `Variable 'Chem.A' not found in variable dictionary` to 5 passing
assertions, agreeing with Rust assertion for assertion — and two it broke.

### 1. An `operator_compose` renaming merge hid the state a test names

A renaming match (§4.7.1 step 4, issue #230) deletes the spelling the test keys
on — `Sink.O3` folded onto `Chem.ozone` — and only a *document* build ever sees
the coupling that does it.
`tests/conformance/merged_rename_reach/fixtures/inline_test_names_the_merged_away_state.esm`
went from 1 pass to `Variable 'O3' not found on compiled system`. The Rust CLI
passes that fixture, so the branch as first pushed would have put Julia and Rust
in disagreement about a document in this repository. The flatten already records
the survivor in `merged_variable_renames`; `_resolve_handle` now follows it,
which is what the tree-walk engine's `_scope_to_component` has always done.

### 2. A spatial sibling took the whole document with it — left as-is, deliberately

`ModelingToolkit.System` refuses a flattened system with spatial independent
variables and redirects to `PDESystem`, which this MTK engine has no route to.
Spatiality is a property of the whole flatten, so one component's
`grad(P, dim: x)` refuses *every* container in that document. On
`tests/valid/units_dimensional_analysis.esm` — five models, and the `grad` is in
`FluidMechanics` alone — that is 13 passing assertions down to 0.

A per-container fallback for such a document was written and then **reverted on
the owner's ruling**: the Rust CLI reports 0 pass / 18 err on the same file,
refusing the whole document for the same `grad`, and a document must not
validate under one binding and fail under another — even when the divergence
predates the change and even when the binding that loses is the one behaving
better. The behaviour is pinned in `container_in_document_test.jl` against the
Rust number, so a future divergence goes red instead of silent.

### 3. An inline test can now assert a variable defined by an equation

`run_inline_tests` resolved an assertion's variable against the ODE state vector
alone (`_scalar_slot`), so every assertion on a variable defined by an algebraic
equation died as `scalar state '<name>' not found`. esm-spec §6.6 names no such
restriction and the Rust CLI returns those values through `observed_field`.
`_observed_field` already treats a shapeless observed as rank 0 precisely so a
contracted scalar is readable; this is a second entry to it, not a second
evaluator.

Closing that closed the second half of the `rhs_time_derivative` category's
Julia exclusion with it — the refusal fixtures' `D` now survives into the
observed body this branch evaluates, where the rewrite-target gate raises the
`unlowered_operator` §4.2 requires. The manifest's `scope_excluded.julia` asked
to be removed "once `run_inline_tests` can read a 0-D observed and refuses an
unlowered `D`"; both hold against every fixture and case, so `julia` joins
`bindings_required` and `rhs_time_derivative_resolution_test.jl` part 2 drives
the manifest here as the Python and Rust runners already do.

### Measured, whole-file

`run_file_tests!` (the ModelingToolkit engine) per file, in a worktree with only
`src/run_tests.jl` swapped between runs. Files not listed were identical across
all columns.

| file | pre-PR | branch as first pushed | now | Rust `esm test` |
|---|---|---|---|---|
| `tests/conformance/rhs_time_derivative/fixtures/tendency_resolution.esm` | 0 pass / 1 err | 5 / 0 | 5 / 0 | 5 / 0 |
| `tests/conformance/merged_rename_reach/fixtures/inline_test_names_the_merged_away_state.esm` | 1 / 0 | **0 / 1** | 1 / 0 | 1 / 0 |
| `tests/valid/units_dimensional_analysis.esm` | 13 / 1 | 0 / 5 | 0 / 5 | 0 / 18 |
| `tests/valid/advection_reaction_loaded_ic_bc.esm` | 0 / 9 | 0 / 1 | 0 / 1 | 0 / 9 |
| `tests/valid/wildfire_atmosphere_ocean.esm` | 0 / 1 | 0 / 1 | 0 / 1 | — |

`run_inline_tests` (the tree-walk engine) before and after the scalar-observed
fix:

| file | before | after | Rust `esm test` |
|---|---|---|---|
| `tests/valid/units_conversions.esm` | 0 / 28 err | **28 / 0** | 28 / 0 |
| `tests/valid/units_propagation.esm` | 0 / 10 err | **10 / 0** | 0 / 10 |
| `tests/conformance/rhs_time_derivative/fixtures/tendency_resolution.esm` | 3 / 2 err | **5 / 0** | 5 / 0 |
| `test/fixtures/container_in_document/surviving_template_ref.esm` | 0 / 1 err | **1 / 0** | 1 / 0 |
| `test/fixtures/container_in_document/sibling_species_read.esm` | 0 / 1 err | **1 / 0** | 1 / 0 |
| `test/fixtures/container_in_document/sibling_species_tendency.esm` | 0 / 3 err | **3 / 0** | 3 / 0 |
| `test/fixtures/container_in_document/sibling_species_system_alone.esm` | 1 / 0 | 1 / 0 | 1 / 0 |
| `tests/valid/units_dimensional_analysis.esm` | 0 / 18 | 0 / 18 | 0 / 18 |
| `tests/simulation/mathematical_correctness.esm` | 33 / 0 | 33 / 0 | 33 / 0 |
| `tests/valid/tests_analyses_comprehensive.esm` | 11 / 0 | 11 / 0 | 11 / 0 |

All 23 Julia test files that drive either inline-test runner pass. The only one
that went red on the way — `rhs_time_derivative_resolution_test.jl`, on the two
testsets that existed to assert Julia's own exclusion — went red exactly where
it says it should ("when either stops being true this test fails and Julia
should join `bindings_required`"), and was rewritten to pin the new behaviour
rather than adjusted to fit.

### Gaps found and deliberately NOT fixed

- **`tests/valid/units_dimensional_analysis.esm` under Rust.** 0 pass / 18 err,
  because one component's unlowered `grad` refuses the whole document. Julia now
  does the same. Whether an unlowerable sibling *should* refuse a document's
  other containers is a question for all five bindings, not a Julia patch.
- **Nothing in continuous integration runs the Julia ModelingToolkit inline-test
  runner over `tests/`.** The Julia suite points `run_esm_tests` only at fixture
  directories under `pkg/EarthSciAST.jl/test/fixtures/` — `inline_tests`, and now
  `container_in_document`. No job exercises it against the shared `tests/` tree.
  That is why this branch was green in continuous integration while losing
  assertions on two documents that live in `tests/`. No sweep job is being built
  for it here; the two testsets added to `container_in_document_test.jl` name
  those two documents directly instead.

### 4. A zero-length time span is answered, not integrated (Rust)

Fixing (3) left a window: `tests/valid/units_propagation.esm` passed 10 under
Julia and Python and 0 under Rust, which refused all ten with `diffsol error:
ODE solver error: Stop time is at the current state time`. Closed here rather
than accepted.

**Which bug it is.** Not "the time span should never have been zero-length".
esm-spec §6.6.2 names the shape — "the instantaneous-derivative test shape
(observed tendencies asserted at `time: 0`)" — §6.6 constrains an assertion's
`time` only to lie in `[time_span.start, time_span.end]`, which `0` does in
`[0, 0]`, and the schema's `TimeSpan` carries no `end > start` rule. The
document is in `tests/valid/`, 23 corpus documents author the shape, and two of
the three bindings that execute inline tests already evaluated it. The bug is
that Rust asked diffsol to step to a stop time it was already at instead of
returning the initial state.

**Where.** `run_solver` (`simulate/driver.rs`) is the only caller of
`set_stop_time` in the crate, and both the scalar backend
(`simulate/compiled.rs`) and the array backend (`simulate_array/driver.rs`)
route through it — so one guard covers both. It sits beside the branch that
already answered requested times at or before `t0` from the initial state; this
is the same answer for the whole grid when the run cannot get past `t0`.
`set_stop_time` moved a few lines down and still runs on exactly the intervals
it always accepted, so nothing about driving diffsol over a non-empty interval
changes.

**The class, swept.** 45 tests across 23 documents declare
`time_span.start == time_span.end`. Twenty-two of those documents are STATIC —
no ODEs, so `solve` returns `NotDynamic` before reaching a solver and the runner
answers from the build — which is why the shape looked supported. Measured with
`esm test` on binaries built either side of the change, **exactly one row
moves**:

| document | Rust before | Rust after |
|---|---|---|
| `tests/valid/units_propagation.esm` | 0 pass / 10 err | **10 / 0** |
| the other 22 zero-length-span documents | unchanged | unchanged |

The two that still report an error do so on purpose:
`precision/f32_mixed_element_types.esm` (`the_document_must_not_build`) and
`undeclared_operand/undeclared_operand_in_max.esm`.

Pinned by `a_zero_length_time_span_answers_from_the_initial_state`, which
asserts a state and an observed of an *integrating* component over `{0, 0}`,
the same over `{2, 2}` so the guard cannot be read as a `t == 0` special case,
and the moving `{0, 1}` control beside them. `cargo test --all-features`:
148 suites, 1647 tests, 0 failures.

### Where the two executing bindings now stand

Every document measured during this review, Julia's tree-walk runner against the
Rust command-line tool, at the tip of this branch:

| document | Julia | Rust |
|---|---|---|
| `tests/valid/units_conversions.esm` | 28 / 0 | 28 / 0 |
| `tests/valid/units_propagation.esm` | 10 / 0 | 10 / 0 |
| `tests/valid/units_dimensional_analysis.esm` | 0 / 18 | 0 / 18 |
| `tests/valid/tests_analyses_comprehensive.esm` | 11 / 0 | 11 / 0 |
| `tests/simulation/mathematical_correctness.esm` | 33 / 0 | 33 / 0 |
| `tests/conformance/rhs_time_derivative/fixtures/tendency_resolution.esm` | 5 / 0 | 5 / 0 |
| `tests/conformance/merged_rename_reach/fixtures/inline_test_names_the_merged_away_state.esm` | 1 / 0 | 1 / 0 |
| `test/fixtures/container_in_document/surviving_template_ref.esm` | 1 / 0 | 1 / 0 |
| `test/fixtures/container_in_document/sibling_species_read.esm` | 1 / 0 | 1 / 0 |
| `test/fixtures/container_in_document/sibling_species_tendency.esm` | 3 / 0 | 3 / 0 |
| `test/fixtures/container_in_document/sibling_species_system_alone.esm` | 1 / 0 | 1 / 0 |

Eleven of eleven identical. Six of those rows disagreed when this review
started.

### DIFF
--- a/pkg/EarthSciAST.jl/src/run_tests.jl
+++ b/pkg/EarthSciAST.jl/src/run_tests.jl
@@ -612,6 +612,33 @@ function _compile_model(model, name::Symbol)
     return MTK.mtkcompile(sys)
 end
 
+# Build a model container from the WHOLE-DOCUMENT `FlattenedSystem` rather than
+# from the container in isolation.
+#
+# `MTK.System(::Model)` wraps the one model in a SYNTHETIC single-model
+# `EsmFile` (`flatten(::Model)`) and flattens that, which silently drops
+# everything the document supplies beyond the container itself:
+#
+#   * the document's `expression_templates` registry — `Model` has no such
+#     field, the registry is document-scoped — so `expand_flattened_refs` sees
+#     an EMPTY registry, becomes a no-op, and a §9.6.4 Option-B
+#     `apply_expression_template` reference survives into `_esm_to_symbolic`
+#     as `Unsupported operator: apply_expression_template`;
+#   * every SIBLING component's variables, so an equation reading another
+#     component's state (`D(Rs.A, t)`, or a plain `Rs.A`) dies in
+#     `_resolve_lowering_var` as `Variable 'Rs.A' not found in variable
+#     dictionary`.
+#
+# Both are one bug: the container is not the document. The Python gate builds
+# `esm_problem` over `flatten(ef)` and the Rust CLI builds over the whole
+# flattened document too (§6.6 selects WHICH tests run, not what the system
+# contains), so this path was the 1-of-3 outlier among the executing runners.
+function _compile_model_in_document(flat, name::Symbol)
+    MTK = _require_mtk()
+    sys = MTK.System(flat; name=name)
+    return MTK.mtkcompile(sys)
+end
+
 function _compile_reaction_system(rs, name::Symbol)
     MTK = _require_mtk()
     cat = _try_require(_CATALYST_PKGID)
@@ -696,10 +723,39 @@ function run_file_tests!(results::Vector{AssertionResult}, path::AbstractString;
     # so it applies to every container in the file.
     stiffness = esm_file.solver === nothing ? nothing : esm_file.solver.stiffness
 
+    # The document flatten every in-document container is built from, computed
+    # at most ONCE per file and only when a container that needs it actually
+    # has tests (a container with no tests is never compiled, so it is never
+    # lowered — the §9.5.3a refusal stays attached to a build that happens).
+    # Table lowering runs at FILE scope here because the flatten happens once,
+    # ahead of the per-container builds that would otherwise each lower their
+    # own.
+    doc_flat = nothing
+    doc_flat_err = nothing
+
     if esm_file.models !== nothing
+        if any(m -> !isempty(m.tests), values(esm_file.models))
+            try
+                lower_table_lookups!(esm_file)
+                doc_flat = flatten(esm_file;
+                                   base_path=dirname(abspath(String(path))))
+            catch err
+                doc_flat_err = err
+            end
+        end
         for (mname, model) in esm_file.models
+            if doc_flat_err !== nothing
+                for t in model.tests
+                    push!(results, AssertionResult(
+                        path, :model, String(mname), t.id, 0, "", NaN, NaN,
+                        nothing, ERROR,
+                        "Model compile failed: $(doc_flat_err)", 0.0))
+                end
+                continue
+            end
             _run_container_tests!(results, path, :model, String(mname), model,
-                                  _compile_model, "Model";
+                                  (_c, sym) -> _compile_model_in_document(doc_flat, sym),
+                                  "Model";
                                   function_tables=tables,
                                   stiff_files=stiff_files, stiffness=stiffness,
                                   solver_hints=esm_file.solver)
@@ -707,13 +763,48 @@ function run_file_tests!(results::Vector{AssertionResult}, path::AbstractString;
     end
 
     if esm_file.reaction_systems !== nothing
+        # A mechanism that shares its document with a model may read that
+        # model — RADM2's 156 reaction rates are scoped references to the
+        # component that evaluates its rate coefficients — and its tests may
+        # override that model's parameters. The isolated Catalyst build cannot
+        # see it: the species resolve, but the sibling's parameters are not on
+        # the compiled system, so the SOLVE fails rather than the build. Such a
+        # document is built from the same whole-document flatten the models use
+        # (§7.4 lowers the mechanism to its mass-action ODEs there).
+        #
+        # A document whose ONLY containers are reaction systems keeps the
+        # Catalyst route: it is the one that carries Catalyst's own default
+        # seeding, and it is the build the corpus's large mechanisms have
+        # always taken.
+        mixed = esm_file.models !== nothing && !isempty(esm_file.models)
         for (rname, rs) in esm_file.reaction_systems
-            _run_container_tests!(results, path, :reaction_system,
-                                  String(rname), rs, _compile_reaction_system,
-                                  "ReactionSystem"; esm_container=rs,
-                                  function_tables=tables,
-                                  stiff_files=stiff_files, stiffness=stiffness,
-                                  solver_hints=esm_file.solver)
+            if mixed && doc_flat_err === nothing && !isempty(rs.tests)
+                if doc_flat === nothing
+                    try
+                        lower_table_lookups!(esm_file)
+                        doc_flat = flatten(esm_file;
+                                           base_path=dirname(abspath(String(path))))
+                    catch err
+                        doc_flat_err = err
+                    end
+                end
+            end
+            if mixed && doc_flat_err === nothing && doc_flat !== nothing
+                _run_container_tests!(results, path, :reaction_system,
+                                      String(rname), rs,
+                                      (_c, sym) -> _compile_model_in_document(doc_flat, sym),
+                                      "ReactionSystem";
+                                      function_tables=tables,
+                                      stiff_files=stiff_files, stiffness=stiffness,
+                                      solver_hints=esm_file.solver)
+            else
+                _run_container_tests!(results, path, :reaction_system,
+                                      String(rname), rs, _compile_reaction_system,
+                                      "ReactionSystem"; esm_container=rs,
+                                      function_tables=tables,
+                                      stiff_files=stiff_files, stiffness=stiffness,
+                                      solver_hints=esm_file.solver)
+            end
         end
     end
 end
--- a/pkg/EarthSciAST.jl/test/container_in_document_test.jl
+++ b/pkg/EarthSciAST.jl/test/container_in_document_test.jl
@@ -1,0 +1,54 @@
+# A container's inline tests build over the WHOLE DOCUMENT, not over the
+# container in isolation (src/run_tests.jl).
+#
+# `MTK.System(::Model)` wraps one model in a synthetic single-model `EsmFile`,
+# which drops everything the document supplies around that container. Two
+# things the corpus actually relies on went with it:
+#
+#   * the document-scoped `expression_templates` registry (`Model` carries no
+#     such field), so `expand_flattened_refs` ran against an EMPTY registry,
+#     silently did nothing, and a §9.6.4 Option-B `apply_expression_template`
+#     reference reached the symbolic lowering as
+#     `Unsupported operator: apply_expression_template`;
+#   * every sibling component's variables, so an equation reading another
+#     component — `D(Rs.A, t)`, the §6.6 instantaneous-derivative shape for a
+#     mechanism's species tendencies, or a plain `Rs.A` — died as
+#     `Variable 'Rs.A' not found in variable dictionary`.
+#
+# The Python gate builds `esm_problem` over `flatten(ef)` and the Rust CLI
+# builds over the whole flattened document; §6.6 selects WHICH tests run, not
+# what the system contains. These fixtures are the two symptoms and a control.
+using Test
+using EarthSciAST
+import ModelingToolkit
+import Catalyst
+import OrdinaryDiffEqTsit5
+import OrdinaryDiffEqRosenbrock
+import OrdinaryDiffEqNonlinearSolve
+
+const _cid_dir = joinpath(@__DIR__, "fixtures", "container_in_document")
+
+@testset "a container builds over its document" begin
+    results, exit_code = run_esm_tests([_cid_dir]; root=_cid_dir, verbose=false)
+
+    @test exit_code == 0
+    @test !isempty(results)
+    @test all(r -> r.status == EarthSciAST.PASS, results)
+
+    # Name the two symptoms explicitly: a regression in either one reappears as
+    # its own message, and a bare "some assertion failed" would not say which.
+    for r in results
+        @test !occursin("Unsupported operator: apply_expression_template",
+                        something(r.message, ""))
+        @test !occursin("not found in variable dictionary",
+                        something(r.message, ""))
+    end
+
+    # The product-only species P is reached through the same path as the
+    # substrates: it has a net stoichiometry, so it has an ODE, and its
+    # tendency is asserted alongside theirs.
+    ids = Set(r.test_id for r in results)
+    @test "tendencies_at_t0" in ids
+    @test "template_reference_at_t0" in ids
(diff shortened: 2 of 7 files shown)


ctessum-claude:
Test status, as promised in the description.

**CI: 20/20 green**, including `julia-tests` on 1.10, 1.12 and 1.13 (the only skip is `Dependency Security Analysis`, which skips on this repo generally).

**Local `Pkg.test()` on 1.12: 25571 pass / 7 broken / 25578 total, exit 0**, 56m19s. The 7 broken are the suite's pre-existing `@test_broken` entries, not this change.

The new test's own evidence, since a green suite alone does not show a regression test bites: `container_in_document_test.jl` is **16/16 with this change and 7 pass / 5 fail without it** (source change stashed, fixtures and test left in place). The 5 failures are the two symptoms by name — `Unsupported operator: apply_expression_template` and `Variable 'Rs.A' not found in variable dictionary` — which is what the per-message assertions in that file are for.

Issue #277, opened 2026-09-08
§9.6.3 constraint 6 pre-evaluation walk exists only in Python (and Rust); Julia/Go/TS gate reactively


ctessum-claude:
esm-spec §9.6.3 constraint 6 makes the unlowered-operator gate a **pre-evaluation walk**:

> "before a component is EVALUATED or COMPILED for simulation, its expression trees are walked; any node whose `op` is not in the evaluable-core set … is rejected."

#237 implemented that walk in **Python only** (`_assert_no_unlowered_operator`, at the `esm_problem` front door). The other bindings that evaluate still need review. Investigation results from that work, recorded so they aren't lost:

## Per-binding state

**Rust** — already has a system-level walk (`first_unlowered_operator` / `reject_unlowered_operators`, `flatten.rs:934/950`), called from both simulators' compile guards. Likely already conformant. Needs confirmation that it covers the **dead-observed** case and `field_ics`.

**Julia** — has the gap, in a worse form than Python did. `_fold_elementwise_array_observeds` (`build_helpers.jl:564`) **deletes dead observeds outright** before any walk could see them. A walk must therefore go in `_build_evaluator_impl_inner` *before* `_build_lower_and_classify` (`build.jl:3336`), not after.

**Go and TypeScript** — both ship public evaluators, so neither is exempt under the "parse/validate-only tooling" clause. Both gate reactively and both leak through lazy `ifelse`/`and`/`or` branches.

> **Hazard for TypeScript:** its gate tests `getOpInfo(op) === undefined`, and the TS `OPS` registry contains only scalar ops. A naive whole-tree walk there would wrongly reject `aggregate`, `makearray`, `index` and the rest of the core array vocabulary. The evaluable-core set must be sourced from something other than `OPS`.

## Why it matters

The asymmetry is not cosmetic. In Python, before #237, the scalar pathway lambdified every observed eagerly and so tripped over an unlowered op in a *dead* observed, while the array pathway evaluated lazily and never reached one — so the array pathway silently under-enforced a normative gate. Any binding that evaluates lazily has the same exposure.

## Related

- #237 (Python implementation, and the two `constrained_match_scope` fixtures it newly rejects)
- #274 (a different §9.5.3 gap on `esm_problem`'s `FlattenedSystem` carrier)

Found during subagent review of #237.

Issue #247, opened 2026-09-08
esm-spec: "an unevaluable op is a diagnostic, never a number" is a format invariant that nothing normative states, and five bindings spell it five ways


ctessum:
## Summary

All five bindings enforce the invariant #220 names — *"an unevaluable op is a diagnostic, never a number"* — but **nothing normative says they must**, and each spells the diagnostic differently. The audit below was produced while fixing #220 (PR #245), which closed the one real hole: Rust's scalar interpreter returned `NaN` where its array interpreter reported.

#220 framed the invariant as a property of the **format**. It is currently a property of five independent implementation choices that happen to agree.

## The audit

| Binding | Terminal arm for an op with no rule |
|---|---|
| Rust | `unevaluable_operator` (both evaluators, after #245) |
| Julia | `throw(TreeWalkError("E_TREEWALK_UNSUPPORTED_OP", op))` in all three walkers; `ArgumentError` on the MTK path |
| Python | `raise NumpyInterpreterError` / `raise SimulationError` — no code |
| TypeScript | `throw UnloweredOperatorError` / `EvaluatorError('unsupported_operator')` |
| Go | `unevaluableOpError`, which already classifies closed-core vs open-tier |

Five spellings for one condition. A consumer cannot detect "this document used an op the runtime cannot evaluate" portably.

## Four secondary gaps found in the same audit

None returns a number, so none is the #220 bug — but each weakens the diagnostic:

- **Julia** — `skolem` and `apply_expression_template` get past the *named* build gate via `_cse_hoistable`, surfacing later under a generic code instead of the specific one.
- **Python** — the terminal raise carries no diagnostic code at all, and there is no build-time gate; the failure appears at evaluation.
- **TypeScript** — reports eight of #220's nine ops as open-tier *rewrite targets*, which is wrong advice: they are closed-core ops that an earlier pipeline stage must eliminate. It also refuses `{"op":"true"}` although fixtures ship it.
- **Go** — omits `distinct` from `closedNonScalarOps`, and raises `enum` untyped.

## Suggested scope

1. State the invariant normatively in `esm-spec.md`: an operator with no evaluation rule MUST be reported, and MUST NOT evaluate to a sentinel value.
2. Adopt one diagnostic code across the bindings for it.
3. Close the four gaps above.
[...]

PR #337, opened 2026-09-15
fix(julia,rust): the §9.6.3 unlowered-operator walk reaches trees the build discards

ctessum-claude:
## What was wrong

esm-spec §9.6.3 constraint 6 makes the `unlowered_operator` gate a pre-evaluation **walk**: every expression tree of a component is walked before it is evaluated or compiled, and any op outside the evaluable core is refused. #237 made Python's check a real walk. This PR closes the remaining gaps in the two bindings that have a build step.

**Julia.** The tree-walk compiler refuses an unlowered op in every tree it compiles, but the build discards some trees first: a dead observed (nothing reads it) is never compiled, and `_fold_elementwise_array_observeds` folds an elementwise array observed into its readers and drops it. Reproduction on `main` (`esm_problem(doc, (0.0, 1.0))`):

| document | before |
|---|---|
| dead scalar observed `extra = div(c)` | builds, no error |
| dead scalar observed `extra = godunov_hamiltonian(c)` | builds, no error |
| dead array observed `extra = div(c) + 1` (folded away) | builds, no error |
| dead array observed `extra = div(c)` | `E_TREEWALK_UNSUPPORTED_SHAPE` (wrong code) |
| live RHS `ifelse(c > 100, godunov_hamiltonian(c), 1)` | `unlowered_operator` (already correct) |

Python and Rust refuse all of the first four.

**Rust.** The scalar interpreter and the single-model array entry already walk everything they build (dead observeds and untaken branches included; verified). One hole remained: `flatten` routes every `ic(state) = rhs` equation into `FlattenedSystem::field_ics`, and `ArrayCompiled::from_flattened` builds its synthetic model from the equations alone, so the stage-(0) walk never saw an initial-condition right-hand side. `ic(c) = godunov_hamiltonian(2.0)` compiled, and failed only at `solve` with "Invalid field initial condition … RHS is neither a provider-served loaded field, a constant, nor a per-cell coordinate expression".

## What changed

- **Julia** — new `src/tree_walk/unlowered_gate.jl`: `_reject_unlowered_operators(model)` walks every equation (and subsystem) before any pass that drops a tree, at both build entry points (`_build_evaluator_impl_inner`, right after template expansion; and the `build_evaluator(::AbstractDict)` front door, ahead of the binning-coordinate / value-invention pre-passes that evaluate observed bodies). A `D` is core only on an equation LHS with a time `wrt` (so a `D` under an LHS `faq` still builds); anything else outside the §4.2 core is refused with `unlowered_operator`, naming the op. The core set is the op registry plus the core ops with no registry row (`table_lookup`, `apply_expression_template`, `rank`, `distinct`, `argmin`, `argmax`). It gates ops, not deadness: a dead observed whose body is fully lowered still builds (CONFORMANCE_SPEC §5.27.3).
- **Rust** — `ArrayCompiled::from_flattened` runs `check_evaluable` over `flat.field_ics` at build.

## Behaviour change (disclosed)

A Julia document with an unlowered op in a dead or folded-away observed, or in an initial condition, used to build and now fails at build with `unlowered_operator`. This converges with Python (#237), Rust, and the spec. A Rust document on the flattened array path with an unlowered op in a field initial condition now fails at build with `unlowered_operator` instead of at `solve` with a generic error.

## Relationship to the other bindings

- **Python** — already done in #237.
- **Go and TypeScript** — their public scalar evaluators gate lazily, so an unlowered op in an untaken `ifelse` branch evaluates to a number (reproduced on `main`: `Evaluate(ifelse(u > 100, grad(p), 1))` → `1` in both). That is being fixed by the `unevaluable_operator` work for #247/#281, whose pre-evaluation walk refuses both `unlowered_operator` and `unevaluable_operator` in one pass (untaken branches included). Adding a second walk here would duplicate it, so this PR does not touch Go or TypeScript, and is **Part of #277** rather than closing it: #277 closes when both land.

## Tests

Each regression test was run on the unfixed code first and failed, then passed with the fix.

- **Julia** `test/unlowered_operator_walk_test.jl` (new): 15/15 with the fix; on unfixed code 11 of the first 13 assertions failed (the initial-condition case was added afterwards). The two controls, a fully lowered dead observed and a structural `D` under an LHS `faq`, still build.
- **Rust** `tests/field_ic_unlowered_operator.rs` (new): 2/2 with the fix; on unfixed code the refusal test failed (the document compiled) and the constant-initial-condition control passed.
- **Julia, existing files** (minimal environment, per file): `rhs_time_derivative_resolution_test.jl`, `expression_templates_test.jl`, `pde_inline_dead_observed_test.jl`, `scope_injection_test.jl`, `template_imports_test.jl`, `build_inspection_test.jl`, `error_hierarchy_test.jl`, `deprecated_op_alias_test.jl`, `flatten_conformance_test.jl`, `structural_validation_test.jl`, `inline_tests_test.jl`, `loaded_ic_bc_simulation_test.jl`, `simulate_run_test.jl`, `wildfire_simulation_test.jl`, `test_file_prelude_test.jl` all pass; `broadcast_alignment_test.jl` 264/264; `simulate_e2e_test.jl` 48/48. `tree_walk_test.jl` 149/150: the one failure is the `t_build < 5.0` timing assertion on the 1024-state diagonal system, which measures first-call compilation on a shared, loaded machine. Timed directly, a second build of that model and the walk alone each finish well inside the limit, so the walk is not the cause.
- **Rust, existing targets**: `loaded_ic_bc_simulation`, `coupled_array_seam`, `discrete_materialize_conformance`, `evaluable_core_gate`, `simulate` all pass. `cargo fmt --all -- --check` clean. `cargo clippy --all-targets -- -D warnings` is clean on every target except one lint already on `main` and untouched here (`clippy::collapsible_match` at `src/flatten.rs:838`, reported by the local 1.95 toolchain; CI pins 1.98), so it was re-run with only that lint allowed: clean.
- **Cross-language conformance** `./scripts/test-conformance.sh --skip-binding-suites`: exit 0, 40 stages passed, none failed.

Part of #277.


ctessum-claude:
Cross-reference: the Go and TypeScript half of #277 (a pre-evaluation walk that refuses `unlowered_operator` in untaken branches) landed in #335 (issue #247/#281), not in this PR. #277 can be closed once both #335 and this PR are merged.

